PR #3, opened 2020-06-18
Add operator configmap into shared_info

dantingl:
Node controller gets operator configmap from shared_info, or gets it from API call.


dantingl:
Old branch and repo are gone, will re-push soon.

PR #21, opened 2020-06-29
Retrieve lb-secret to generate operator configmap

dantingl:


pkg/controller/configmap/configmap_controller.go
```diff
@@ -175,10 +175,18 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			}
 			agentConfigMap = nil
 		}
+		lbSecret := &corev1.Secret{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: ncptypes.LbSecret}, lbSecret)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				log.Error(err, "Failed to get lb-secret")
```
ggverma, line 182: is it ok to proceed further if err is not nil?

dantingl: If operator failed to get lb-secret, I think we can let operator re-create the secret again. Do you think it's ok?

dantingl:
Original branch and repo got deleted, will re-push the patch

PR #22, opened 2020-06-30
Apply config if NCP image changes

dantingl:


pkg/controller/configmap/configmap_controller.go
```diff
@@ -190,7 +192,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	if !ncpNeedChange && !agentNeedChange {
+	// Check if NCP_IMAGE changes
+	ncpImageChanged, err := r.isNcpImageChanged()
+	if err != nil {
+		return reconcile.Result{}, err
```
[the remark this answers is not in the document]
salv-orlando: Good point, it looks like the error for deployment retrieval in isNcpImageChanged can only be temporary

pkg/controller/configmap/configmap_controller.go
```diff
@@ -190,11 +192,22 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
 	if !ncpNeedChange && !agentNeedChange {
```
ggverma, line 196: I suppose this is to update the running NCP image. However, since we watch for changes in only nsx-ncp operator configmap and network CRD as per https://github.com/vmware/nsx-container-plugin-operator/blob/754394ce588b0e3fe24d23756a9de4103a4c4bbb/pkg/controller/configmap/configmap_controller.go#L98 in configmap_controller, wouldn't we require an update in them to trigger this? Not sure if it is expected...

dantingl: As NCP_IMAGE is in operator.yaml as an environment variable, only operator restart can trigger the image change. So each time when operator restarts, it can get an update from both operator configmap and network CRD. Then operator can compare with the NCP_IMAGE at that time.

ggverma: so one caveat is that `!ncpNeedChange && !agentNeedChange` is true. wonder if we can support live updates and if it's worth it

dantingl: If we want to support live update, then we could move the env variable to operator configmap for NCP image, it is fine for me. I guess the image mostly is changed for NCP upgrade case, maybe not happen frequently?

ggverma: yeah, it likely won't. we should probably check how operator handles/requires upgrades as that's the end goal. lgtm for now

PR #25, opened 2020-07-01
Add operator configmap into sharedInfo

dantingl:
Node controller will use NSX config from operator configmap

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -13,7 +13,10 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/vmware/go-vmware-nsxt"
+	nsxt "github.com/vmware/go-vmware-nsxt"
+	"github.com/vmware/nsx-container-plugin-operator/pkg/controller/sharedinfo"
+	"github.com/vmware/nsx-container-plugin-operator/pkg/controller/statusmanager"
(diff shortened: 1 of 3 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -163,14 +165,19 @@ func filterSegmentPorts(nsxClients *NsxClients, ports []*data.StructValue, nodeN
...
-	log.Info("Getting NCP configmap for node controller")
-	ncpConfigMap := &corev1.ConfigMap{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: operatortypes.NsxNamespace, Name: operatortypes.NcpConfigMapName}, ncpConfigMap)
-	if err != nil {
-		return nil, err
+	configMap := r.sharedInfo.OperatorConfigMap
+	if configMap == nil {
+		log.Info("Getting config from operator configmap")
+		configMap = &corev1.ConfigMap{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: operatortypes.OperatorNamespace, Name: operatortypes.ConfigMapName}, configMap)
```
ggverma, line 172: nit: rename `operatortypes` to `ncptypes` or vice-versa to avoid confusion

dantingl: Sure, I noticed that too. Will unify to operatortypes in a following patch.

PR #27, opened 2020-07-02
Support NSX cert/key/ca

heypnus:
1. Support the user configure NSX cert/key/ca in operator
configmap yaml file
2. Node controller can connect to NSX manager via NSX cert/key

### DIFF
--- a/manifest/cluster-network-15-nsx-secret.yaml
+++ b/manifest/cluster-network-15-nsx-secret.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-data: {tls.crt: null, tls.key: null}
+data: {tls.crt: "{{.NsxCert}}", tls.key: "{{.NsxKey}}", tls.ca: "{{.NsxCA}}"}
 kind: Secret
 metadata: {name: nsx-secret, namespace: nsx-system}
 type: kubernetes.io/tls
--- a/manifest/cluster-network-19-nsx-ncp.yaml
+++ b/manifest/cluster-network-19-nsx-ncp.yaml
@@ -53,3 +53,9 @@ spec:
               - {key: tls.crt, path: lb-cert/tls.crt}
               - {key: tls.key, path: lb-cert/tls.key}
               name: lb-secret
+          - secret:
+              items:
+              - {key: tls.crt, path: nsx-cert/tls.crt}
+              - {key: tls.key, path: nsx-cert/tls.key}
+              - {key: tls.ca, path: nsx-cert/tls.ca}
+              name: nsx-secret
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -9,6 +9,9 @@ import (
 	"strings"
 
 	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/pkg/errors"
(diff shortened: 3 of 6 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -187,24 +220,73 @@ func (r *ReconcileNode) createNsxClients() (*NsxClients, error) {
...
+			securityCtx.SetProperty(security.PASSWORD_KEY, password)
+			nsxClients.PolicySecurityContext = securityCtx
+			log.Info("Using username and password to connect to NSX")
+		}
+	}
+
 	// policy client
-	tlsConfig := tls.Config{InsecureSkipVerify: true}
-	tlsConfig.BuildNameToCertificate()
+	tlsConfig, err := getConnectorTLSConfig(insecure, tmpCertPath, tmpKeyPath, tmpCAPath)
```
ggverma, line 282: Is the `err` handled correctly here?

pkg/controller/node/node_controller.go
```diff
@@ -187,24 +220,73 @@ func (r *ReconcileNode) createNsxClients() (*NsxClients, error) {
...
+		insecure = true
+	}
 	nsxClients := NsxClients{}
 	nsxClients.Cluster = cluster
 
+	// Get cert/key/ca, then write then into temp files
+	var certData = cfg.Section("operator").Key("nsx_api_cert").Value()
+	var keyData = cfg.Section("operator").Key("nsx_api_private_key").Value()
+	var caData = cfg.Section("operator").Key("nsx_ca").Value()
+	certDataDecoded, err := base64.StdEncoding.DecodeString(certData)
```
ggverma, line 234: nit: 234-245, 255-266 can make use of a func

manifest/cluster-network-15-nsx-secret.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-data: {tls.crt: null, tls.key: null}
+data: {tls.crt: "{{.NsxCert}}", tls.key: "{{.NsxKey}}", tls.ca: "{{.NsxCA}}"}
```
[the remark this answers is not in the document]
dantingl: Yes, I think so.

pkg/controller/node/node_controller.go
```diff
@@ -164,6 +167,36 @@ func filterSegmentPorts(nsxClients *NsxClients, ports []*data.StructValue, nodeN
...
 
+func getConnectorTLSConfig(insecure bool, clientCertFile string, clientKeyFile string, caFile string) (*tls.Config, error) {
+	tlsConfig := tls.Config{InsecureSkipVerify: insecure}
+
+	if len(clientCertFile) > 0 {
+		if len(clientKeyFile) == 0 {
+			return nil, fmt.Errorf("Please provide key file for client certificate")
+		}
+
+		cert, err := tls.LoadX509KeyPair(clientCertFile, clientKeyFile)
```
dantingl, line 178: I found there is another function  X509KeyPair to use PEM encoded data directly, maybe we don't need to save to a file in the middle?

manifest/cluster-network-15-nsx-secret.yaml
dantingl, line 2: Right now even operator puts CA cert in nsx-secret, but how does NCP read it? I think we lost the implementation on NCP side to read CA from secret, not sure how does QE test it before.

pkg/controller/node/node_controller.go
heypnus: Done

pkg/controller/node/node_controller.go
heypnus: Good catch, done, thanks.

manifest/cluster-network-15-nsx-secret.yaml
dantingl: Thanks, maybe later we can update this to README.

manifest/cluster-network-19-nsx-ncp.yaml
```diff
@@ -53,3 +53,9 @@ spec:
               - {key: tls.crt, path: lb-cert/tls.crt}
               - {key: tls.key, path: lb-cert/tls.key}
               name: lb-secret
+          - secret:
```
dantingl, line 56: Looks like we will need to update ncp.yaml later in NCP repo.

dantingl:
Later you should do the same like PR #24 to check if nsx-secret changes.

PR #31, opened 2020-07-03
Fix port not found issue when node name doesn't match vSphere name

heypnus:
Node controller will get the providerID from OpenShift node.spec,
then use the providerID to achieve the vSphere VM name which is needed
when searching node port.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -110,7 +110,29 @@ type NsxClients struct {
 
 var cachedNodeSet = map[string](*statusmanager.NodeStatus){}
 
-func searchSegmentPortByNodeNameAddress(nsxClients *NsxClients, nodeName string, nodeAddress string) (*model.SegmentPort, error) {
+func getVcNameByProviderId(nsxClients *NsxClients, nodeName string, providerId string) (string, error) {
+	// providerId has the following format: vsphere://<uuid>
+	if len(providerId) != 46 {
+		return "", errors.Errorf("Invalid provider ID %s of node %s", providerId, nodeName)
+	}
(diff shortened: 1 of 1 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -110,7 +110,30 @@ type NsxClients struct {
...
+	for _, vm := range(vms.Results) {
+		for _, computeId := range(vm.ComputeIds) {
+			// format of computeId: biosUuid:<uuid>
+			if providerId == string([]byte(computeId)[9:]) {
+				return vm.DisplayName, nil
+			}
+		}
+	}
+	return "", errors.Errorf("No virtual machine matches provider ID %s and hostname %s", providerId, nodeName)
+
```
dantingl, line 133: Remove empty line 133

pkg/controller/node/node_controller.go
```diff
@@ -413,7 +436,18 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
 
 	cluster := nsxClients.Cluster
-	segmentPort, err := searchSegmentPortByNodeNameAddress(nsxClients, nodeName, nodeAddress)
+	providerId := instance.Spec.ProviderID
+	nodeVcName, err := getVcNameByProviderId(nsxClients, nodeName, providerId)
+	if err != nil {
+		cachedNodeSet[nodeName] = &statusmanager.NodeStatus{
+			Address: nodeAddress,
+			Success: false,
+			Reason:  fmt.Sprintf("Error while achieving  vsphere name for node %s: %v", nodeName, err),
```
dantingl, line 445: Extra whitespace in 'achieving  vsphere'

PR #33, opened 2020-07-06
Ensure service CIDRs are updated in any case

salv-orlando:
As NCP does not handle service CIDRs, if these are the only
item being updated on the network CR, NCP skips procesing
the resource status update.

This patch ensures the network CR is updated if necessary
even if the operator is not applying any change to NCP.

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -196,6 +196,10 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			}
 		}
 	}
+	networkConfigChanged := true
+	if r.sharedInfo.NetworkConfig != nil {
+		networkConfigChanged = HasNetworkConfigChange(networkConfig, r.sharedInfo.NetworkConfig)
+	}
 	ncpNeedChange, agentNeedChange, err := NeedApplyChange(instance, appliedConfigMap)
 	if err != nil {
 		return reconcile.Result{}, err
@@ -210,6 +214,15 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 
 		if !ncpImageChanged {
 			log.Info("no new configuration needs to apply")
+			// Check if network config status must be updated
+			if networkConfigChanged {
+				err = updateNetworkStatus(networkConfig, r)
+				if err != nil {
+					r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
+						fmt.Sprintf("Failed to update network status: %v", err))
+				}
+				return reconcile.Result{}, err
+			}
 			r.status.SetNotDegraded(statusmanager.ClusterConfig)
 			r.status.SetNotDegraded(statusmanager.OperatorConfig)
 			return reconcile.Result{}, nil
@@ -272,11 +285,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	r.sharedInfo.OperatorConfigMap = appliedConfigMap
 
 	// Update network CRD status
-	err = updateNetworkStatus(networkConfig, r)
-	if err != nil {
-		r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
-			fmt.Sprintf("Failed to update network status: %v", err))
-		return reconcile.Result{}, err
+	if networkConfigChanged {
+		err = updateNetworkStatus(networkConfig, r)
+		if err != nil {
+			r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
+				fmt.Sprintf("Failed to update network status: %v", err))
+			return reconcile.Result{}, err
+		}
 	}
 
 	r.status.SetNotDegraded(statusmanager.ClusterConfig)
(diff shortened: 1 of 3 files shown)


pkg/controller/configmap/configmap_controller.go
```diff
@@ -196,6 +196,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			}
 		}
 	}
+    networkConfigChanged := true
+    if r.sharedInfo.NetworkConfig != nil {
+        networkConfigChanged = HasNetworkConfigChange(networkConfig, r.sharedInfo.NetworkConfig)
```
ggverma, line 201: nit: `HasNetworkConfigChange` -> `hasNetworkConfigChanged`?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -196,6 +196,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
+        if err != nil {
```
ggverma, line 202: where do we set `err `? might be irrelevant in context with prev comment

pkg/controller/configmap/configmap_controller.go
```diff
@@ -196,6 +196,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			}
 		}
 	}
+    networkConfigChanged := true
```
ggverma, line 199: is it simpler if we can do:
`networkConfigChanged := hasNetworkConfigChanged(networkConfig, r.sharedInfo.NetworkConfig)`

pkg/controller/configmap/configmap_controller.go
```diff
@@ -210,6 +217,15 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 
 		if !ncpImageChanged {
 			log.Info("no new configuration needs to apply")
+            // Check if network config status must be updated
```
ggverma, line 220: is the indentation correct here?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -272,20 +288,22 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	r.sharedInfo.OperatorConfigMap = appliedConfigMap
 
 	// Update network CRD status
-	err = updateNetworkStatus(networkConfig, r)
-	if err != nil {
-		r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
-			fmt.Sprintf("Failed to update network status: %v", err))
-		return reconcile.Result{}, err
-	}
+    if networkConfigChanged {
```
ggverma, line 291: indent here as well

pkg/controller/configmap/configmap_controller.go
salv-orlando: makes sense. I think I had something different in mind initially!

pkg/controller/configmap/configmap_controller.go
salv-orlando: I'll verify on the source code.

pkg/controller/configmap/configmap_controller.go
salv-orlando: You're right but I kept it capital for consistency with NeedAppluChange

pkg/controller/configmap/configmap_controller.go
salv-orlando: Indentation is correct but my editor is using spaces instead of tabs... hence the problem.
Let me fix it.

pkg/controller/configmap/configmap_controller.go
salv-orlando: Ah I did set it to trye here to ensure it was defined even if NetworkConfig was not yet populated (first reconcilation)

pkg/controller/configmap/configmap_controller.go
salv-orlando: initially I assumed HasNetworkConfigChange would return an error, but now it's clear I can simply remove this code.

pkg/controller/configmap/configmap_controller.go
salv-orlando: removed

pkg/controller/configmap/configmap_controller.go
salv-orlando: indent seems ok on last PR update

pkg/controller/configmap/configmap_controller.go
salv-orlando: fixed as well

pkg/controller/configmap/configmap_controller.go
```diff
@@ -204,12 +211,17 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	if !ncpNeedChange && !agentNeedChange {
 		// Check if NCP_IMAGE changes
 		ncpImageChanged, err := r.isNcpImageChanged()
-		if err != nil {
```
ggverma, line 207: why don't we need this? perhaps you removed the wrong `err` check with line 202?

salv-orlando: meh I was doing too many things at the same time

ggverma: thought so

PR #34, opened 2020-07-07
Do not use golang base image

dantingl:
Dockerfile.1 image: 685MB
Dockerfile.2 image: 668MB
Dockerfile.3 image: 258MB


dantingl:
Will use another way to build the binary.

PR #35, opened 2020-07-08
Build binary from Makefile

dantingl:
Remove golang middle image from Dockerfile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,26 @@
+# go options
+GO                 ?= go
+LDFLAGS            :=
+GOFLAGS            :=
+BINDIR             ?= $(CURDIR)/build/bin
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+all: build
+.PHONY: all
+
+LDFLAGS += $(VERSION_LDFLAGS)
+
+.PHONY: build
+build:
+	GOOS=linux $(GO) build -o $(BINDIR)/nsx-ncp-operator $(GOFLAGS) -ldflags '$(LDFLAGS)' ./cmd/manager
+	docker build -f build/Dockerfile . -t nsx-ncp-operator
```
dantingl, line 16: Better to separate these two commands

PR #38, opened 2020-07-20
Add image location and cleanup configmap

yasensim:
@salv-orlando can I ask you to review this pr?

### DIFF
--- a/deploy/configmap.yaml
+++ b/deploy/configmap.yaml
@@ -8,23 +8,13 @@ data:
 
     [vc]
 
-    # IpAddress or Hostname of VC
-    #vc_endpoint = <None>
-
-    # The SSO domain associated with the deployment
-    #sso_domain = vsphere.local
-
-    # VC API server HTTPS port.
-    #https_port = 443
-
-
     [coe]
 
     # Container orchestrator adaptor to plug in.
     adaptor = openshift4
 
     # Specify cluster for adaptor.
-    cluster = openshift
+    # cluster = {cluster_name}
 
     # Log level for NCP operations
     # Choices: NOTSET DEBUG INFO WARNING ERROR CRITICAL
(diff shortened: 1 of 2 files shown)


yasensim:
@dantingl  could you also review it please?

yasensim:
The idea of this pr is to cleanup parameters that are not relevant to Openshift in the configmap

deploy/configmap.yaml
```diff
@@ -8,23 +8,13 @@ data:
 
     [vc]
```
dantingl, line 9: We can remove the section header [vc] as well

deploy/configmap.yaml
```diff
@@ -540,44 +468,23 @@ data:
...
-
-
-
-    #[nsx_v3]
-    # Deprecated option: tier0_router
-    # Replaced by [nsx_v3] top_tier_router
-
-    # Deprecated option: deny_subnets_redistribution
-    # Replaced by [nsx_v3] configure_t0_redistribution
+    process_oc_network = False
```
salv-orlando, line 478: We should be default to False here in the operand, aren't we?

deploy/operator.yaml
```diff
@@ -23,7 +23,7 @@ spec:
       containers:
         - name: nsx-ncp-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: vmware/nsx-container-plugin-operator:0.0.3
```
salv-orlando, line 26: We have now 0.0.4 and soon 0.0.5, perhaps we should update this PR

deploy/configmap.yaml
```diff
@@ -87,18 +73,11 @@ data:
     # Total number of compressed backup files to store. Defaults to 5.
     #log_rotation_backup_count = 5
 
-
-
-
     [nsx_v3]
```
salv-orlando, line 76: I think there are some topology related options that can be dropped

salv-orlando:
I believe this PR is now superseded, therefore closing it. Reopen it if necessary.

PR #40, opened 2020-07-24
Add Operator CRD

dantingl:
Operator should own a CRD, add NCP replicas as the property in the CRD


dantingl:
> Do we need to list the CRD in the role binding (role.yaml) ?

Sure, will add it in role binding

pkg/controller/configmap/config.go
```diff
@@ -208,9 +208,13 @@ func Render(configmap *corev1.ConfigMap) ([]*unstructured.Unstructured, error) {
 		}
 	}
 	if haEnabled {
-		renderData.Data[operatortypes.NcpReplicasKey] = operatortypes.NcpHaReplicas
+		renderData.Data[operatortypes.NcpReplicasKey] = ncpReplicas
 	} else {
 		renderData.Data[operatortypes.NcpReplicasKey] = 1
+		if ncpReplicas != 1 {
+			log.Info(fmt.Sprintf("Set nsx-ncp deployment replicas to 1 instead of %d as HA is disabled",
```
ggverma, line 215: this looks like a warning log. too bad this logger doesn't have it. maybe we can use its verbosity feature if we stick to this

pkg/controller/configmap/configmap_controller.go
```diff
@@ -101,17 +108,37 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
 	}
 
+	// Fetch ncp-install CRD instance
+	ncpInstallCrd := &operatorv1.NcpInstall{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: operatortypes.NcpInstallCRDName,
+		Namespace: operatortypes.OperatorNamespace}, ncpInstallCrd)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info(fmt.Sprintf("%s CRD is not found", operatortypes.NcpInstallCRDName))
+			r.status.SetDegraded(statusmanager.OperatorConfig, "NoNcpInstallCRD",
```
ggverma, line 124: `"NoNcpInstallCRD"` should be a const here https://github.com/vmware/nsx-container-plugin-operator/blob/0471eb4eb93e11570cca8a9212e4e9182c51aceb/pkg/controller/statusmanager/status_manager.go#L31

pkg/controller/configmap/configmap_controller.go
```diff
@@ -101,17 +108,37 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
+				fmt.Sprintf("%s CRD is not found", operatortypes.NcpInstallCRDName))
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		r.status.SetDegraded(statusmanager.OperatorConfig, "NoNcpInstallCRD",
+			fmt.Sprintf("Failed to get operator CRD: %v", err))
+		return reconcile.Result{}, err
```
ggverma, line 131: not requeueing but mentioned in the comment

pkg/controller/configmap/configmap_controller.go
```diff
@@ -101,17 +108,37 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
+		Namespace: operatortypes.OperatorNamespace}, ncpInstallCrd)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info(fmt.Sprintf("%s CRD is not found", operatortypes.NcpInstallCRDName))
+			r.status.SetDegraded(statusmanager.OperatorConfig, "NoNcpInstallCRD",
+				fmt.Sprintf("%s CRD is not found", operatortypes.NcpInstallCRDName))
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		r.status.SetDegraded(statusmanager.OperatorConfig, "NoNcpInstallCRD",
```
ggverma, line 129: `"NoNcpInstallCRD"` should be `"InvalidNcpInstallCRD"`? it's present and thus not the same error

pkg/controller/configmap/configmap_controller.go
```diff
@@ -101,17 +108,37 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
+			fmt.Sprintf("Failed to get operator CRD: %v", err))
+		return reconcile.Result{}, err
+	}
+	ncpReplicas := ncpInstallCrd.Spec.NcpReplicas
```
ggverma, line 133: could this be empty or null?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -221,13 +248,13 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 		return reconcile.Result{}, err
 	}
 	if !ncpNeedChange && !agentNeedChange {
-		// Check if NCP_IMAGE changes
-		ncpImageChanged, err := r.isNcpImageChanged()
+		// Check if NCP_IMAGE and nsx-ncp replicas change
```
ggverma, line 251: `Check if NCP_IMAGE and nsx-ncp replicas change` -> `Check if NCP_IMAGE or nsx-ncp replicas change` or just remove the comment

pkg/controller/configmap/configmap_controller.go
ggverma, line 129: do we need to unset `"NoNcpInstallCRD"`?

pkg/controller/configmap/config.go
dantingl: Sure, I'm also thinking to use the log level in our code. How about I update all logs in a following patch.

pkg/controller/configmap/configmap_controller.go
dantingl: It is in OperatorConfig status level, NoNcpInstallCRD is just the reason.

pkg/controller/configmap/configmap_controller.go
dantingl: Sure

dantingl: It should be done in line 327

pkg/controller/configmap/configmap_controller.go
dantingl: Looks like configmap and network CRD have the same issue :)

pkg/controller/configmap/configmap_controller.go
dantingl: Based on my test, the default value is 0. So looks like I need to handle the unset case

pkg/controller/configmap/configmap_controller.go
dantingl: Sure

pkg/controller/configmap/config.go
ggverma: sounds good

pkg/controller/configmap/configmap_controller.go
ggverma: I see, thanks Danting! :)

pkg/controller/configmap/configmap_controller.go
ggverma: oh yeah, integers in go are 0 by default

dantingl:
> is this for K8s? Sorry, I am not sure I understand why we need this CRD

For the certification, operator should own a CRD and update status on the CRD.

PR #50, opened 2020-09-01
Enable Grouping Calculation Notif feature by default

seanpang-vmware:
enable grouping feature in openshift by default


salv-orlando (approved):

pkg/controller/configmap/config.go
```diff
@@ -39,6 +39,7 @@ func FillDefaults(configmap *corev1.ConfigMap, spec *configv1.NetworkSpec) error
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "adaptor", "openshift4", true))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "policy_nsxapi", "True", true))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "single_tier_topology", "True", true))
+	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "wait_for_security_policy_sync", "True", true))
```
dantingl, line 42: Do we need to make it as force value? I thought last 'true' should be 'false'

PR #51, opened 2020-09-02
Support tier1-segment port as the node port

heypnus:
Tier1-segment means the segment who has path
`/infra/tier-1s/<tier-1-id>/segments/<segment-id>`.

Previously we had already supported the infra-segment,
whose path is` /infra/segments/<segment-id>`.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -25,7 +25,8 @@ import (
 	vspherelog "github.com/vmware/vsphere-automation-sdk-go/runtime/log"
 	policyclient "github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/security"
-	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/segments"
+	infra_segment "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/segments"
+	tier1_segment "github.com/vmware/vsphere-automation-sdk-go/services/nsxt/infra/tier_1s/segments"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
 	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/search"
 	"gopkg.in/ini.v1"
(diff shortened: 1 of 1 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -461,8 +462,15 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
 	// TODO: find a way to fill the segmentPort by using the return value from searchNodePortByVcNameAddress then we won't need to invoke the Get API again.
 	sl := strings.Split(*segmentPort.Path, "/")
 	segmentId := sl[len(sl)-1]
-	portsClient := segments.NewDefaultPortsClient(nsxClients.PolicyConnector)
-	*segmentPort, err = portsClient.Get(segmentId, *segmentPort.Id)
+	portsClient := infra_segment.NewDefaultPortsClient(nsxClients.PolicyConnector)
```
dantingl, line 465: Move line 465 to 472?
Please add the path pattern for infra segment in comments as well

heypnus: > Move line 465 to 472?
Previous this line locates at 472, but it reported an error that portsClient is undefined, so I moved it to 465
> Please add the path pattern for infra segment in comments as well
Done

dantingl: Move 465 to 472, and in 470 define portsClient as well
portsClient := tier1_segment.NewDefaultPortsClient(nsxClients.PolicyConnector)

pkg/controller/node/node_controller.go
```diff
@@ -503,7 +515,11 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
 		var clusterTag = model.Tag{Scope: &clusterScope, Tag: &cluster}
 		segmentPort.Tags = append(segmentPort.Tags, clusterTag)
 	}
-	_, err = portsClient.Update(segmentId, *segmentPort.Id, *segmentPort)
```
dantingl, line 506: Sorry I did not know we still need portsClient here

dantingl (approved):

pkg/controller/node/node_controller.go
heypnus: Yes and I found the Update methods from infra_segment and tier1_segment are different so in the latest change I split the clients to infraPortsClient and tier1PortsClient

ggverma (approved):

PR #53, opened 2020-09-09
Set operator image

yasensim:
Set real image to the operator deployment


salv-orlando:
We also need more changes to the CSV in oder to satisfy Operator-SDK scorecard tests. Instead of opening a new PR we will update this one

salv-orlando:
This does not seem to be needed anymore

PR #55, opened 2020-09-15
Support MTU configuration in NCP Operator

dantingl:
Set MTU in network CRD status
Restart nsx-ncp-bootstrap pods if MTU is changed

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -43,8 +43,9 @@ func FillDefaults(configmap *corev1.ConfigMap, spec *configv1.NetworkSpec) error
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "enable_snat", "True", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "ha", "enable", "True", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "k8s", "process_oc_network", "False", true))
-    // For Openshift add a 3-seconds agent delay by default
+	// For Openshift add a 3-seconds agent delay by default
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "waiting_before_cni_response", "3", false))
+	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "mtu", strconv.Itoa(operatortypes.DefaultMTU), false))
 	appendErrorIfNotNil(&errs, fillClusterNetwork(spec, cfg))
 
 	// Write config back to ConfigMap data
@@ -128,6 +129,12 @@ func validateConfigMap(configmap *corev1.ConfigMap) []error {
 	if validateConfig(cfg, "nsx_v3", "tier0_gateway") != nil &&
 		validateConfig(cfg, "nsx_v3", "top_tier_router") != nil {
 		appendErrorIfNotNil(&errs, errors.Errorf("failed to get tier0_gateway or top_tier_router"))
+	}
+	// Check MTU value
+	mtu := cfg.Section("nsx_node_agent").Key("mtu").Value()
+	_, err = strconv.Atoi(mtu)
+	if err != nil {
+		appendErrorIfNotNil(&errs, errors.Wrapf(err, "mtu is invalid"))
 	}
 
 	return errs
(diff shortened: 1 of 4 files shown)


pkg/controller/configmap/config.go
```diff
@@ -43,7 +43,7 @@ func FillDefaults(configmap *corev1.ConfigMap, spec *configv1.NetworkSpec) error
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "enable_snat", "True", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "ha", "enable", "True", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "k8s", "process_oc_network", "False", true))
-    // For Openshift add a 3-seconds agent delay by default
+	// For Openshift add a 3-seconds agent delay by default
```
salv-orlando, line 46: thanks! I think I authored that patch with the editor with the python settings...

pkg/controller/configmap/config.go
```diff
@@ -266,26 +270,28 @@ func HasNetworkConfigChange(currConfig *configv1.Network, prevConfig *configv1.N
 	return stringSliceEqual(currCidrs, prevCidrs)
 }
 
-func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap) (ncpNeedChange bool, agentNeedChange bool, err error) {
+func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap) (
+	ncpNeedChange bool, agentNeedChange bool, bootstrapNeedChange bool, err error) {
```
salv-orlando, line 274: do you think there might be a case for creating a struct describing which components need a configuration change?

pkg/controller/configmap/config.go
```diff
@@ -339,6 +354,15 @@ func inSlice(str string, s []string) bool {
 	return false
 }
 
+func inMapKey(str string, m map[string][]string) bool {
```
salv-orlando, line 357: I might be wrong but perhaps there's an easier way for this?

_val, found := m[key]

source: https://blog.golang.org/maps

pkg/controller/configmap/config.go
```diff
@@ -431,3 +455,29 @@ func optionInConfigMap(configMap *corev1.ConfigMap, section string, key string)
...
+	}
+	if sec.HasKey(key) {
+		return sec.Key(key).Value()
+	}
+	return ""
+}
+
+func IsMTUChanged(currConfigMap *corev1.ConfigMap, prevConfigMap *corev1.ConfigMap) bool {
+	currMtu := getOptionInConfigMap(currConfigMap, "nsx_node_agent", "mtu")
+	prevMtu := getOptionInConfigMap(prevConfigMap, "nsx_node_agent", "mtu")
```
salv-orlando, line 477: just to make sure I read the code correctly, these will always be strings right?

pkg/controller/configmap/config.go
dantingl: I added a map for bootstrap pod in pkg/types/defaults.go, do you mean to add a whole list for all components?

pkg/controller/configmap/config.go
dantingl: n_n👍

pkg/controller/configmap/config.go
dantingl: Yes, values are all strings
https://godoc.org/gopkg.in/go-ini/ini.v1#Key.Value

pkg/controller/configmap/config.go
salv-orlando: I did not mean the section.
I notice that the function is now return a tuple of 4 values, and I was wondering it might be more manageable to have the 3 boolean in a struct

pkg/controller/configmap/config.go
salv-orlando: ok, we're good then ;)

pkg/controller/configmap/config.go
dantingl: OK, sure

pkg/controller/configmap/config.go
```diff
@@ -310,24 +322,34 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
...
+		}
+		if !needChange.agent && inSlice(sec, operatortypes.AgentSections) {
+			needChange.agent = true
 		}
-		if !agentNeedChange && inSlice(sec, operatortypes.AgentSections) {
-			agentNeedChange = true
+		bootstrapOpts, found := operatortypes.BootstrapOptions[sec]
+		if !needChange.bootstrap && found {
+			for _, opt := range bootstrapOpts {
+				if currCfg.Section(sec).Key(opt).Value() != prevCfg.Section(sec).Key(opt).Value() {
```
ggverma, line 337: could there be an error here if the `Key` is not found in prev or curr cfg?

pkg/controller/configmap/config.go
```diff
@@ -431,3 +453,29 @@ func optionInConfigMap(configMap *corev1.ConfigMap, section string, key string)
...
+	if err != nil {
+		return ""
+	}
+	if sec.HasKey(key) {
+		return sec.Key(key).Value()
+	}
+	return ""
+}
+
+func IsMTUChanged(currConfigMap *corev1.ConfigMap, prevConfigMap *corev1.ConfigMap) bool {
```
ggverma, line 473: not necessary but, do you think we can be proactive and validate new MTU size before updating configmap and restarting pods?

pkg/controller/configmap/config.go
dantingl: If Key is not found, it will return a zero-value. 
Refer to https://godoc.org/gopkg.in/go-ini/ini.v1#Section.Key

pkg/controller/configmap/config.go
dantingl: Good suggestion, we can add MTU validation in operator

salv-orlando (approved):

pkg/controller/configmap/config.go
```diff
@@ -431,3 +456,29 @@ func optionInConfigMap(configMap *corev1.ConfigMap, section string, key string)
...
+
+func getOptionInConfigMap(configMap *corev1.ConfigMap, section string, key string) string {
+	cfg, err := ini.Load([]byte(configMap.Data[operatortypes.ConfigMapDataKey]))
+	if err != nil {
+		log.Error(err, "Failed to load ConfigMap")
+		return ""
+	}
+	sec, err := cfg.GetSection(section)
+	if err != nil {
+		return ""
```
heypnus, line 468: better log error before return ""?

heypnus (approved):

PR #56, opened 2020-09-16
Final CSV and support for custom Namespace

salv-orlando:
This patch updates the CSV for the operator by:
- adding NcpInstalls CRD to permissions
- ensuring OLM will create ClusterRole for permissions
- Fixing various typos
- Adding descriptors for CRD spec and status fields
- Adjusting install modes to ensure only Own and Single Namespace
  are supported
- ensuring hostNetwork permissions are given
- adding WATCH_NAMESPACE env variable to the deployment spec

The CSV version is therefore bumped to 0.0.4
A python utility to create the bundle zip for the latest CSV has
been added (make_zip_bundle.py)

In additioni, the operator code has been modified to support the
SingleNamespace install mode and ensuring it is not executed in
Multi or AllNamespace mode (in both cases it will default to
nsx-system-operator)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,14 +9,17 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"github.com/vmware/nsx-container-plugin-operator/pkg/apis"
 	"github.com/vmware/nsx-container-plugin-operator/pkg/controller"
+	operatortypes "github.com/vmware/nsx-container-plugin-operator/pkg/types"
 	"github.com/vmware/nsx-container-plugin-operator/version"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
@@ -59,6 +62,35 @@ func main() {
 
 	printVersion()
 
+	// Check if we need to watch a specific namespace
+	namespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		log.Error(err, "Failed to get watch namespace")
+		os.Exit(1)
+	}
+
+	if strings.Contains(namespace, ",") {
+		namespace = ""
+		log.Info(`This operator cannot handle multiple
+			  namespaces. The operator will watch for changes
+		          across all namespaces`)
+	}
+
+	// Set default manager options
+	options := manager.Options{
+		Namespace: namespace,
+	}
+
+	if namespace != "" && namespace != operatortypes.NsxNamespace {
+		// since the pod controller must watch nsx-system namespace,
+		// this becomes akin to a multinamespace scenario, but we
+		// will treat it as "AllNamespaces" since the MultiNamespaceCache
+		// does not load cluster-scoped resources.
+		// The namespace will be loaded anyway into the statusmanager so
+		// the configMapController will monitor only that namespace
+		options.Namespace = ""
+	}
+
 	// Get a config to talk to the apiserver
 	cfg, err := config.GetConfig()
 	if err != nil {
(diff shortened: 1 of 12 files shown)


bundle/make_zip_bundle.py
```diff
@@ -0,0 +1,55 @@
...
+    import yaml
+except ImportError:
+    print("Please install pyyaml")
+    sys.exit(1)
+
+import zipfile
+
+
+def parse_package_manifest(manifest_path):
+    f = open(manifest_path)
```
ggverma, line 15: perhaps put the file open in a `with` block and read inside `try`?

ggverma: also, do we need to verify that file exists?

cmd/manager/main.go
```diff
@@ -59,6 +62,26 @@ func main() {
...
+		namespace = ""
+		log.Info(`This operator cannot handle multiple
+			  namespaces. The operator will watch for changes
+		          across all namespaces`)
+	}
+
+	if namespace == "" {
+		namespace = operatortypes.OperatorNamespace
+		log.Info(`This operator must watch resources in a single
+                          namespace. Setting default namespace value`)
```
ggverma, line 82: it might be helpful to log the the namespace value itself

ggverma: also, does this mean that we won't reconcile changes in pod controller since it watches resources in nsx-system ns?

salv-orlando: hmm, I think I need to make it work since it will expect now that the pod and daemonset are in the same namespace as the operator

cmd/manager/main.go
```diff
@@ -59,6 +63,33 @@ func main() {
...
+			  namespaces. The operator will watch for changes
+		          across all namespaces`)
+	}
+
+	// Set default manager options
+	options := manager.Options{
+		Namespace: namespace,
+	}
+
+	if namespace != "" && namespace != operatortypes.NsxNamespace {
```
ggverma, line 84: so if the WATCH_NAMESPACE is "nsx-system", configmap_controller won't work?

ggverma (approved):

cmd/manager/main.go
salv-orlando: In this case namespace value will be set to nsx-system and Options.Namespace will be nsx-system
this means all the controllers will monitor exclusively nsx-system which should be fine, as in this case the operator and operand namespace are the same

pkg/controller/configmap/configmap_controller.go
```diff
@@ -133,7 +137,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	// Fetch ncp-install CRD instance
 	ncpInstallCrd := &operatorv1.NcpInstall{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: operatortypes.NcpInstallCRDName,
-		Namespace: operatortypes.OperatorNamespace}, ncpInstallCrd)
+		Namespace: r.status.OperatorNamespace}, ncpInstallCrd)
```
dantingl, line 140: Same here, if WATCH_NAMESPACE is set as "", do we need to use "nsx-system-operator" to get resources?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -109,9 +109,13 @@ type ReconcileConfigMap struct {
...
-
-	// Check request namespace and name to ignore other changes
-	if request.Namespace == operatortypes.OperatorNamespace {
+	watchedNamespace := r.status.OperatorNamespace
+	if watchedNamespace == "" {
+		log.Info(fmt.Sprintf("ConfigMapController can only watch a single namespace, defaulting to: %s",
+			operatortypes.OperatorNamespace))
+		watchedNamespace = operatortypes.OperatorNamespace
+	}
+	if request.Namespace == r.status.OperatorNamespace {
```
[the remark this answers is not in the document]
salv-orlando: ah good point, I forgot to add that ;)

pkg/controller/configmap/configmap_controller.go
salv-orlando: yup

dantingl (approved):

pkg/controller/configmap/configmap_controller.go
salv-orlando: I modified the code ensured everything works properly if the operator is deployed in `nsx-system` namespace

PR #60, opened 2020-09-28
Add nsx-secret in shared_info

dantingl:
node_controller will use nsx-secret from shared_info

### DIFF
--- a/pkg/controller/sharedinfo/shared_info.go
+++ b/pkg/controller/sharedinfo/shared_info.go
@@ -17,6 +17,7 @@ type SharedInfo struct {
 
 	NetworkConfig     *configv1.Network
 	OperatorConfigMap *corev1.ConfigMap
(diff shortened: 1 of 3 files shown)


pkg/controller/sharedinfo/shared_info.go
```diff
@@ -17,6 +17,7 @@ type SharedInfo struct {
 
 	NetworkConfig     *configv1.Network
 	OperatorConfigMap *corev1.ConfigMap
+	OperatorNsxSecret *corev1.Secret
```
heypnus, line 20: Should we also add lb secret into the shared info?

dantingl: Resources in shared_info are shared between controllers. As we don't have a controller using lb secret, I think we don't need add it there.

heypnus (approved):

PR #64, opened 2020-10-19
Search node LSP by provider ID

heypnus:
To avoid tagging wrong port when two ports have the same vm-name and the
same IP addresses:
1. use manager API ListVirtualMachines to get the VM by providerID and
   get the VM's property "externalId"
2. use manager API ListVifs to get the Vif by externalId then get the
   Vif's property "lport_attachment_id"
3. use manager API ListLogicalPorts to get the logical switch port by
   lport_attachment_id
4. get the correct port from the results of 3 according to the address


dantingl (approved):
LGTM

PR #66, opened 2020-11-23
Support generic NCP operator

heypnus:
This patch will allow the user deploy NCP/nsx-node-agent on
native Kubernetes by the operator


deploy/kubernetes/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
   namespace: nsx-system-operator
 spec:
   ncpReplicas: 1
+  disableTag: true
```
dantingl, line 8: How about changing disableTag to addNodeTag?

manifest/rhel_on_kubernetes/cluster-network-03-nsxerrors.nsx.vmware.com.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: apiextensions.k8s.io/v1
```
dantingl, line 1: How about adding subdir /rhel and /ubuntu under manifest/kubernetes/

olm-catalog/0.0.4/operator.nsx.vmware.com_ncpinstalls_crd.yaml
```diff
@@ -1 +1 @@
-../../deploy/operator.nsx.vmware.com_ncpinstalls_crd.yaml
\ No newline at end of file
+../../deploy/kubernetes/operator.nsx.vmware.com_ncpinstalls_crd.yaml
```
dantingl, line 1: It should be ../../deploy/openshift4/operator.nsx.vmware.com_ncpinstalls_crd.yaml

pkg/controller/configmap/config.go
```diff
@@ -112,7 +66,38 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
 	return nil
 }
 
-func validateConfigMap(configmap *corev1.ConfigMap) []error {
```
dantingl, line 115: Not related to your patch, but do you think we should move all functions in this file to ReconcileConfigMap class? Is that a better style in go?

pkg/controller/configmap/config.go
```diff
@@ -149,7 +134,11 @@ func getPrefixFromCIDR(cidr string) (uint32, error) {
 	return uint32(i), nil
 }
 
-func validateClusterNetwork(spec *configv1.NetworkSpec) []error {
+func (r *ReconcileConfigMap) validateClusterNetwork(spec *configv1.NetworkSpec) []error {
```
dantingl, line 137: I think it's better to split ReconcileConfigMap func and ReconcileConfigMapOc func into 2 config files. And maybe add interfaces for them?

pkg/controller/configmap/config.go
```diff
@@ -256,7 +249,30 @@ func Render(configmap *corev1.ConfigMap, ncpReplicas int32, nsxSecret *corev1.Se
 	return objs, nil
 }
 
-func HasNetworkConfigChange(currConfig *configv1.Network, prevConfig *configv1.Network) bool {
+func (r *ReconcileConfigMap) GetManifestDir() (string, error) {
+	release, err := os.Open("/host/etc/os-release")
```
dantingl, line 253: Add "/host/etc/os-release" as a constant var?

pkg/controller/configmap/config.go
```diff
@@ -256,7 +249,30 @@ func Render(configmap *corev1.ConfigMap, ncpReplicas int32, nsxSecret *corev1.Se
...
+	if err != nil {
+		return "", err
+	}
+	defer release.Close()
+	content, err := ioutil.ReadAll(release)
+	osRelease := string(content)
+	if strings.Contains(osRelease, "CoreOS") {
+		// TODO: add platform detection logic if we allow the user deploy coreos image on native k8s
```
dantingl, line 261: I think QE has test to run CoreOS on K8s, but we should still use rhel_on_kubernetes manifests for it, right?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -41,23 +45,81 @@ var appliedConfigMap *corev1.ConfigMap
...
+	reconcileConfigMap := ReconcileConfigMap{
 		client:     mgr.GetClient(),
 		scheme:     mgr.GetScheme(),
 		status:     status,
 		sharedInfo: sharedInfo,
 	}
+	if sharedInfo.AdaptorName == "openshift4" {
+		return &ReconcileConfigMapOc{reconcileConfigMap}
+	} else {
+		return &ReconcileConfigMapK8s{reconcileConfigMap}
```
dantingl, line 64: ReconcileConfigMapK8s is not used in config.go

pkg/controller/configmap/configmap_controller.go
```diff
@@ -41,23 +45,81 @@ var appliedConfigMap *corev1.ConfigMap
...
+	}
+
+}
+
+func (r *ReconcileConfigMap) FillDefaults(configmap *corev1.ConfigMap, spec *configv1.NetworkSpec) error {
```
dantingl, line 69: Move FillDefaults to config.go

pkg/controller/configmap/configmap_controller.go
```diff
@@ -183,30 +279,32 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: operatortypes.NetworkCRDName}, networkConfig)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			log.Info("Cluster network CRD is not found")
-			r.status.SetDegraded(statusmanager.ClusterConfig, "NoClusterConfig", "Cluster network CRD is not found")
-			return reconcile.Result{}, nil
+	if r.sharedInfo.AdaptorName == "openshift4" {
```
dantingl, line 282: You can use a ReconcileConfigMapOc func to get network CR

pkg/controller/node/node_controller.go
```diff
@@ -451,6 +430,26 @@ func (r *ReconcileNode) createNsxClients() (*NsxClients, error) {
...
 
+func (r *ReconcileNode) isTagDisabled() (bool, error) {
+	ncpInstall := &operatorv1.NcpInstall{}
+	watchedNamespace := r.status.GetOperatorNamespace()
+	if watchedNamespace == "" {
+		log.Info(fmt.Sprintf("NodeController can only watch a single namespace, defaulting to: %s",
+			operatortypes.OperatorNamespace))
+		watchedNamespace = operatortypes.OperatorNamespace
+	}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: operatortypes.NcpInstallCRDName, Namespace: watchedNamespace}, ncpInstall)
```
dantingl, line 441: How about put ncpInstall in sharedInfo, so you don't need to get it again in node_controller

pkg/controller/sharedinfo/shared_info.go
```diff
@@ -4,22 +4,60 @@
...
 	configv1 "github.com/openshift/api/config/v1"
+	operatortypes "github.com/vmware/nsx-container-plugin-operator/pkg/types"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+var log = logf.Log.WithName("controller_configmap")
```
dantingl, line 22: controller_configmap -> shared_info?

pkg/controller/pod/pod_controller.go
```diff
@@ -132,12 +138,20 @@ var _ reconcile.Reconciler = &ReconcilePod{}
 type ReconcilePod struct {
 	client     client.Client
 	scheme     *runtime.Scheme
-	status     *statusmanager.StatusManager
+	status     statusmanager.StatusManagerInterface
```
dantingl, line 141: I'm not sure if it's good to add 'Interface' in name

deploy/kubernetes/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml
heypnus: It looks better, renamed.

manifest/rhel_on_kubernetes/cluster-network-03-nsxerrors.nsx.vmware.com.yaml
heypnus: done

pkg/controller/configmap/config.go
heypnus: I only did the necessary moving (e.g. the logics for k8s and openshift are different)... Also not sure whether what you said is better.

pkg/controller/sharedinfo/shared_info.go
heypnus: done

pkg/controller/node/node_controller.go
heypnus: Yes the code is redundant, will try to improve it.

pkg/controller/configmap/configmap_controller.go
heypnus: Done

pkg/controller/configmap/config.go
heypnus: will try it.

pkg/controller/node/node_controller.go
heypnus: done.

pkg/controller/pod/pod_controller.go
heypnus: Done. The changes:
type StatusManager struct -> type Status struct
type StatusManagerK8s struct -> type StatusK8s struct
type StatusManagerOc struct -> type StatusOc struct
type StatusManagerInterface interface -> type StatusManager interface

README.md
```diff
@@ -5,7 +5,7 @@
 ## Overview
 
 An operator for leveraging NSX as the default container networking solution for an
-Openshift cluster. The operator will be deployed in the early phases of cluster
+Kubernetes/Openshift cluster. The operator will be deployed in the early phases of cluster
```
salv-orlando, line 8: When integrating with Kubernetes, is it mandatory  to deploy it in the first phases of cluster deployment? 
I don't want to nitpick on the README file! If I recall the implementation correctly, we may also be able to deploy the operator on a cluster where NCP is already running, is that correct?

deploy/kubernetes/operator.nsx.vmware.com_ncpinstalls_crd.yaml
```diff
@@ -37,6 +37,9 @@ spec:
               type: integer
               format: int32
               minimum: 1
+            addNodeTag:
+              description: Skip tagging node logical switch ports with node name and cluster
```
salv-orlando, line 41: Probably the description should something like "Tag node logical switch port ... when set to True, skip tagging when set to False". In the way it is written now, it seems we need to set it to True to skip tagging

pkg/controller/configmap/config.go
salv-orlando: The function is already applied to ReconcileConfigMap type. Danting, do you mean doing the same for all functions?

deploy/kubernetes/operator.nsx.vmware.com_ncpinstalls_crd.yaml
heypnus: Yes, I changed the option from "disableTag" to "addNodeTag" but forgot to modify the description.. Thanks!

README.md
heypnus: Yes, in kubernetes, we don't have to deploy the operator so early like openshift. And the operator can start on a cluster where NCP is already running, although it's very possible that the operator recreated NCP pod when it founds the configmap between the operator and the existing NCP are different.

pkg/controller/pod/pod_controller.go
```diff
@@ -132,12 +138,20 @@ var _ reconcile.Reconciler = &ReconcilePod{}
...
 	scheme     *runtime.Scheme
-	status     *statusmanager.StatusManager
+	status     statusmanager.StatusManager
 	sharedInfo *sharedinfo.SharedInfo
 
 	nsxNcpResources []types.NamespacedName
 }
 
+type ReconcilePodK8s struct {
+	ReconcilePod
```
ggverma, line 148: maybe use a pointer here?

pkg/controller/pod/pod_controller.go
```diff
@@ -240,6 +255,19 @@ func (r *ReconcilePod) checkIfK8sResourceExists(
 }
 
 func (r *ReconcilePod) createK8sObject(obj *unstructured.Unstructured) error {
```
ggverma, line 257: this is same as in line 284. we should try to reuse this function there to avoid duplication. same for other instances where we use *OC and *K8s objects. maybe embed *K8s into *OC struct to allow reuse (same in other controllers)? or maybe changing to interface can solve it?

pkg/controller/configmap/config.go
heypnus: kept the original implementation of ReconcileConfigMap and additionally add interface Adaptor for it. But I didn't split ConfigMapK8s and ConfigMapOc which are implementing the Adaptor interface into 2 files. Do you have an example for the best practice?

pkg/controller/configmap/config.go
heypnus: I think we don't need to move all functions in this file to ReconcileConfigMap class if not necessary. e.g. I saw https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/rollback.go didn't do that.

pkg/controller/pod/pod_controller.go
heypnus: ReconcilePodK8s was removed in the new patch set.

pkg/controller/pod/pod_controller.go
heypnus: Removed the duplication, thanks.

pkg/controller/configmap/configmap_controller.go
heypnus: Done, thanks.

pkg/controller/configmap/config.go
heypnus: Done

pkg/controller/configmap/config.go
heypnus: I checked the NCP bootstrap code, the CoreOS related logic is added by Michal and only for openshift support. I double confirmed that with QE. So we don't need to care about CoreOS on K8s.

README.md
heypnus: I improved the description but not sure if it is accurate enough, could you please look at it again?

README.md
```diff
@@ -79,22 +90,36 @@ spec:
...
-In the ConfigMap, some fields are mandatory including `cluster`, `nsx_api_managers`,
+single or two Tiers topology on native Kubernetes.
+
+#### Kubernetes
+
+Some fields are mandatory including `adaptor`, `cluster`, `nsx_api_managers`,
+`container_ip_blocks`, `tier0_gateway`(for single T1 case), `top_tier_router`
+(for single T0 case), `external_ip_pools`(for SNAT mode).. If any of above
+options is not provided in the operator ConfigMap, the operator will fail to
+reconcile configurations, error messageswill be added in ncpinstall nsx-ncp
```
dantingl, line 104: `messageswill` miss a space

pkg/controller/configmap/configmap_controller.go
```diff
@@ -95,12 +105,24 @@ var _ reconcile.Reconciler = &ReconcileConfigMap{}
 
 // ReconcileConfigMap reconciles a ConfigMap object
 type ReconcileConfigMap struct {
-	// This client, initialized using mgr.Client() above, is a split client
```
dantingl, line 98: Removed it by mistake?

pkg/controller/node/node_controller.go
```diff
@@ -460,6 +438,11 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Node")
 
+	if r.sharedInfo.AddNodeTag == false {
```
dantingl, line 441: Check this flag in line 72 or 78 when init node controller or start watching nodes

pkg/controller/sharedinfo/shared_info.go
```diff
@@ -4,22 +4,77 @@
...
+			operatortypes.OperatorNamespace))
+		watchedNamespace = operatortypes.OperatorNamespace
+	}
+	configmapName := types.NamespacedName{
+		Namespace: watchedNamespace,
+		Name:      operatortypes.ConfigMapName,
+	}
+	// r.client.Get cannot work at the stage, the solution of reader referred to
+	// https://github.com/jaegertracing/jaeger-operator/pull/814
+	err := reader.Get(context.TODO(), configmapName, configmap)
```
dantingl, line 52: If user first creates operator deployment then configmap, will operator restart or retry to get configmap?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +64,135 @@ type deploymentState struct {
...
+		}
+		for _, node := range nodes.Items {
+			status.setNodeNetworkUnavailable(node.ObjectMeta.Name, false, "NSXNodeAgent", "Waiting for nsx-node-agent to be created")
+		}
+		return
+	}
+	pod := corev1.Pod{}
+	for _, pod = range pods.Items {
+		containerStatus := corev1.ContainerStatus{}
+		// podStatus := SetFromPodsForNodeCondition{Ready: true}
```
dantingl, line 158: Remove this line?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +64,135 @@ type deploymentState struct {
...
+	defer status.Unlock()
+
+	pods := &corev1.PodList{}
+	err := status.client.List(context.TODO(), pods, client.MatchingLabels{"component": operatortypes.NsxNodeAgentContainerName})
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Error getting %s for node condition", operatortypes.NsxNodeAgentContainerName))
+	} else if len(pods.Items) == 0 {
+		log.Info("nsx-node-agent not found for node condition")
+		nodes := corev1.NodeList{}
+		errNode := status.client.List(context.TODO(), &nodes)
```
dantingl, line 145: How about `errNode` -> err?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +64,135 @@ type deploymentState struct {
...
+		if errNode != nil {
+			log.Error(errNode, "Failed to get nodes for condition updating")
+			return
+		}
+		for _, node := range nodes.Items {
+			status.setNodeNetworkUnavailable(node.ObjectMeta.Name, false, "NSXNodeAgent", "Waiting for nsx-node-agent to be created")
```
dantingl, line 151: Please check node existing status to avoid updating the same condition many times, you can add it in a following patch

README.md
heypnus: Done, thanks.

deploy/kubernetes/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
   namespace: nsx-system-operator
 spec:
   ncpReplicas: 1
+  addNodeTag: true
```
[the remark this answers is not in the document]
heypnus: I prefer to default disable it for k8s, too. But in openshift it's enabled by default... I think the difference is not serious, I'm changing it to false.

manifest/kubernetes/rhel/cluster-network-04-loadbalancers.vmware.com.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
heypnus: I think it's a good point. Wwe don't need cluster-network-xx-xx in native kubernetes and it can simplify the manifests generation. Will try it in the next patch. Thanks!

pkg/apis/operator/v1/ncpinstall_types.go
```diff
@@ -18,6 +18,9 @@ type NcpInstallSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	NcpReplicas int32 `json:"ncpReplicas,omitempty"`
+	// TODO: test the value with type bool
```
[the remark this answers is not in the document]
heypnus: Yes, I added a todo but I have tested it. removing it.

pkg/controller/configmap/configmap_controller.go
heypnus: Yes... Fixed, thanks.

pkg/controller/node/node_controller.go
heypnus: Good idea, done.

pkg/controller/sharedinfo/shared_info.go
heypnus: If the user creates operator deployment without configmap, the operator will report "Failed to get configmap"  and exit, then keep restarting, until the configmap is ready.

pkg/controller/statusmanager/pod_status.go
heypnus: Done, thanks.

pkg/controller/statusmanager/pod_status.go
heypnus: hmm... I saw in status_manager.go, the definition of set() also directly invokes client.Status().Update(). Not sure if the Update() itself can avoid the dup updating, I'm adding a TODO to confirm it.

pkg/controller/statusmanager/pod_status.go
heypnus: Done, thanks.

deploy/kubernetes/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml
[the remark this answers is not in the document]
heypnus: Now in deploy/openshift4/operator.nsx.vmware.com_ncpinstalls_crd.yaml the addNodeTag's default value is true, and in deploy/kubernetes/operator.nsx.vmware.com_ncpinstalls_crd.yaml, its default value is false.

pkg/controller/sharedinfo/shared_info.go
```diff
@@ -4,22 +4,77 @@
...
 
-func New() *SharedInfo {
-	return &SharedInfo{}
+func New(mgr manager.Manager, operatorNamespace string) (*SharedInfo, error) {
+	reader := mgr.GetAPIReader()
+	configmap := &corev1.ConfigMap{}
+	watchedNamespace := operatorNamespace
+	if watchedNamespace == "" {
+		log.Info(fmt.Sprintf("SharedInfo can only check a single namespace, defaulting to: %s",
+			operatortypes.OperatorNamespace))
```
dantingl, line 43: miss a whitespace

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +63,135 @@ type deploymentState struct {
...
+	}
+	condition := corev1.NodeCondition{
+		Type:               "NetworkUnavailable",
+		Status:             networkUnavailableCondition,
+		LastHeartbeatTime:  metav1.Now(),
+		LastTransitionTime: metav1.Now(),
+		Reason:             reason,
+		Message:            message,
+	}
+	SetNodeCondition(&node.Status.Conditions, condition)
```
dantingl, line 123: Looks like you already check existing condition in `SetNodeCondition`, you can leverage a return value from `SetNodeCondition` to avoid updating node condition

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +63,135 @@ type deploymentState struct {
...
+	err = status.client.Status().Update(context.TODO(), node)
+	if err != nil {
+		log.Error(err, "Failed to update node condition")
+	} else {
+		log.Info("Updated node condition")
+	}
+}
+
+// Get the pod status from API server
+func (status *StatusManager) SetFromPodsForNodeCondition() {
```
dantingl, line 134: `SetFromPodsForNodeCondition` -> SetNodeConditionFromPods

pkg/controller/statusmanager/pod_status.go
```diff
@@ -61,17 +63,135 @@ type deploymentState struct {
...
+	status.Lock()
+	defer status.Unlock()
+
+	pods := &corev1.PodList{}
+	err := status.client.List(context.TODO(), pods, client.MatchingLabels{"component": operatortypes.NsxNodeAgentContainerName})
```
dantingl, line 139: Is it possible to avoid extra K8s API call, I mean can we leverage existing pod status analysis from `SetFromPodsForOverall`?

pkg/controller/sharedinfo/shared_info.go
heypnus: talked with Danting offline. won't modify it because the indent is tab, not whilespaces.

pkg/controller/statusmanager/pod_status.go
heypnus: I did a test, whenthe node was updated via `status.client.Status().Update(context.TODO(), node)` with the unchanged condition, the lastTransitionTime in that node condition won't be changed. e.g. in operator log we can see (timestamp 1607070276 indicates UTC time 2020/12/4 08:24:36):
```
{"level":"info","ts":1607070276.3644621,"logger":"status_manager","msg":"Setting status NetworkUnavailable to flase for node wdc-rdops-vm10-dhcp-7-184"}
{"level":"info","ts":1607070276.4882815,"logger":"status_manager","msg":"Updated node condition"}
```
but after that the node condition is still:
```
lastHeartbeatTime: "2020-11-18T07:19:46Z"
lastTransitionTime: "2020-12-04T08:21:50Z"
message: NSX node agent is running
reason: NSXNodeAgentReady
status: "False"
type: NetworkUnavailable
```
So I think the `status.client.Status().Update()` can avoid the duplicate update by itself, we don't need to do more things.

pkg/controller/statusmanager/pod_status.go
heypnus: done

pkg/controller/statusmanager/pod_status.go
heypnus: I'm afraid that we cannot leverage SetFromPodsForOverall because it only got the deployment/ds state, while the deatiled pod state we want is in the pod list and we cannot read them from deployment/ds. e.g. openshift clusteroperator only should know how many node-agents are ready but for NetworkUnavailbe condition we have to know the ready/unready pods' locations. So we have to additionally get the pod list.

dantingl (approved):

pkg/controller/statusmanager/pod_status.go
dantingl: But you have an extra API call, right?

heypnus: Make sense. You remind me it's possible that the update request was sent while the api server didn't modify the timestamp.. Will investigate further and confirm if we need to skip the update from our side, the possible improvement will be in the next patch.

PR #71, opened 2020-12-09
Mount NSX/LB secret for Kubernetes

heypnus:


manifest/kubernetes/ubuntu/cluster-network-29-nsx-ncp.yaml
```diff
@@ -62,3 +62,14 @@ spec:
...
+          - secret:
+              items:
+              - {key: tls.crt, path: lb-cert/tls.crt}
+              - {key: tls.key, path: lb-cert/tls.key}
+              name: lb-secret
+          - secret:
+              items:
+              - {key: tls.crt, path: nsx-cert/tls.crt}
+              - {key: tls.key, path: nsx-cert/tls.key}
+              - {key: tls.ca, path: nsx-cert/tls.ca}
```
dantingl, line 74: I remember right now NCP does not get CA cert from secret, maybe we need some change on NCP side as well.

heypnus: Yes, I confirmed with QE that they are adding the mount by themselves. I think it's not a must-have change from NCP side. And maybe the user wants a more flexible configuration? I mean maybe their secrets are not called "lb-secret" and "nsx-secret"...

heypnus: Sorry I misunderstood your meaning, I thought you meant both lb-secret and nsx-secret... Yes for now NCP doesn't have nsx-cert/tls.ca, we can add it on NCP side.

dantingl (approved):

PR #74, opened 2020-12-14
Fix the logic to check replicas when HA is disabled

heypnus:


pkg/controller/configmap/config.go
```diff
@@ -288,24 +288,24 @@ func getPrefixFromCIDR(cidr string) (uint32, error) {
 }
 
 func Render(configmap *corev1.ConfigMap, ncpReplicas int32, nsxSecret *corev1.Secret,
-	lbSecret *corev1.Secret) ([]*unstructured.Unstructured, error) {
+	lbSecret *corev1.Secret) ([]*unstructured.Unstructured, error, *render.RenderData) {
```
dantingl, line 291: do you think `error` should be in the end?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -256,7 +256,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// Render configurations
-	objs, err := Render(instance, ncpReplicas, opNsxSecret, opLbSecret)
+	objs, err, renderData := Render(instance, ncpReplicas, opNsxSecret, opLbSecret)
```
dantingl, line 259: I think it's better to either passing `&ncpReplicas` to `Render`, or reseting `ncpReplicas` in a separate function.

heypnus: Code is modified and in testing.

pkg/controller/configmap/config.go
heypnus: removed *render.RenderData so we don't need to care it.

salv-orlando (approved):

ggverma (approved):

pkg/controller/configmap/config.go
```diff
@@ -324,12 +324,13 @@ func Render(configmap *corev1.ConfigMap, ncpReplicas int32, nsxSecret *corev1.Se
...
 	}
 	if haEnabled {
-		renderData.Data[operatortypes.NcpReplicasKey] = ncpReplicas
+		renderData.Data[operatortypes.NcpReplicasKey] = *ncpReplicas
 	} else {
-		renderData.Data[operatortypes.NcpReplicasKey] = 1
-		if ncpReplicas != 1 {
+		renderData.Data[operatortypes.NcpReplicasKey] = int32(1)
+		if *ncpReplicas != 1 {
+			*ncpReplicas = 1
```
dantingl, line 331: You should move `*ncpReplicas = 1` after logging

pkg/controller/configmap/config.go
```diff
@@ -324,12 +324,13 @@ func Render(configmap *corev1.ConfigMap, ncpReplicas int32, nsxSecret *corev1.Se
 		}
 	}
 	if haEnabled {
-		renderData.Data[operatortypes.NcpReplicasKey] = ncpReplicas
```
dantingl, line 327: I think the code could change to:

```
if not haEnabled and *ncpReplicas != 1 {
        log.Info(fmt.Sprintf("Set nsx-ncp deployment replicas to 1 instead of %d as HA is disabled",
	        *ncpReplicas))
        *ncpReplicas = 1
}
renderData.Data[operatortypes.NcpReplicasKey] = *ncpReplicas
```

heypnus: No, consider this case: haEnabled=false && ncpReplicas=3
In your above code, the renderData.Data[operatortypes.NcpReplicasKey]  will be 3, it's not expected.

dantingl: I don't understand why it will be 3 not 1?

heypnus: Sorry my bad. You are right. Will improve it in the next patch.

PR #81, opened 2020-12-17
Fix README by deleting unnecessary backquote

yktsubo:
Fix README by deleting unnecessary backquote 

The words, adaptor/cluster/nsx_api_managers/tier0_gateway_top_tier_router/external_ip_pools are not treated as inline code because of unnecessary backquote


heypnus (approved):

heypnus (commented):
Thanks for fixing it

salv-orlando (approved):

PR #88, opened 2021-01-07
Use the automatically generated yaml files for manifest

heypnus:
This patch is syncing the latest NCP yamls to operator,
and merging multiple yaml files to one for every platform.


deploy/kubernetes/role.yaml
```diff
@@ -39,7 +39,7 @@ rules:
   resources: [deployments, endpoints, ingressclasses, nodes, pods/log, replicationcontrollers, secrets]
   verbs: [get, list, watch]
 - apiGroups: [vmware.com]
-  resources: [loadbalancers, loadbalancers/status, nsxlbmonitors, nsxlbmonitors/status]
+  resources: [loadbalancers, loadbalancers/status, nsxlbmonitors, nsxlbmonitors/status, virtualnetworkinterfaces, virtualnetworkinterfaces/status, virtualnetworks, virtualnetworks/status]
```
dantingl, line 42: They are for WCP CRDs, do you think we should add them?

manifest/openshift4/coreos/ncp-openshift4.yaml
```diff
@@ -0,0 +1,1231 @@
+
```
dantingl, line 1: If we only support Openshift on CoreOS, no need to add sub-directory under manifest/openshift4?

deploy/kubernetes/role.yaml
heypnus: In NCP yaml the logic is:
```
    142    {% if nsxapi == "policy" %}
    143      - virtualnetworks
    144      - virtualnetworks/status
    145      - virtualnetworkinterfaces
    146      - virtualnetworkinterfaces/status
    147    {% endif %}
```
Looks that it's not only for WCP.

manifest/openshift4/coreos/ncp-openshift4.yaml
heypnus: Just thought maybe we will support more OS for openshift in the future...

dantingl (approved):

PR #95, opened 2021-01-19
Set latest tag when building operator image

salv-orlando:


salv-orlando:
accidentally opened - closing

PR #96, opened 2021-01-19
Set latest tag when building operator image

salv-orlando:
This will ensure the image built is also pushed to the
target registry


ggverma (approved):

salv-orlando:
trying to refresh PR to latest commit

PR #105, opened 2021-02-16
Prepare assets: Ensure kustomize is dynamically downloaded

### LINKED ISSUE #104: Remove kustomize binary
Currently upload of release assets is done via a kustomize binary included in the repo.
It shouldn't work in this way: kustomize should be dinamically installed by the script

Also, it's never nice to keep binaries in repo

salv-orlando:
This commit removes the 'kustomize' binary from the repo
and ensures get-kustomize.sh always leverages the downloaded
version.

Fixes #104

### DIFF
--- a/hack/get-kustomize.sh
+++ b/hack/get-kustomize.sh
@@ -5,21 +5,35 @@
 
 THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 KUSTOMIZE_VERSION="v3.9.1"
+_BINDIR=$THIS_DIR/.bin
 
 # Check if you have kustomize installed, or download it.
 # please note that different versions of kustomize can give different results
 check_or_install_kustomize() {
     # Check if there is already a kustomize binary in $_BINDIR and if yes, check
     # if the version matches the expected one.
-    local kustomize="$(PATH=$THIS_DIR command -v kustomize)"
+    local kustomize="$(PATH=$_BINDIR command -v kustomize)"
     if [ -x "$kustomize" ]; then
-        >&2 echo "Found "$kustomize" version "`$kustomize version --short`
-        echo $kustomize
-        return 0
+        local kustomize_version="$($kustomize version --short)"
+        # Should work with following styles:
+        #  - kustomize/v3.3.0
+        #  - {kustomize/v3.8.2  2020-08-29T17:44:01Z  }
+        kustomize_version="${kustomize_version##*/}"
+        kustomize_version="${kustomize_version%% *}"
+        if [ "${kustomize_version}" == "${KUSTOMIZE_VERSION}" ]; then
+            # If version is exact match, stop here.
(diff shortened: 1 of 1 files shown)


hack/get-kustomize.sh
```diff
@@ -5,21 +5,24 @@
 
 THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 KUSTOMIZE_VERSION="v3.9.1"
+_BINDIR=$THIS_DIR/.bin
 
 # Check if you have kustomize installed, or download it.
 # please note that different versions of kustomize can give different results
 check_or_install_kustomize() {
     # Check if there is already a kustomize binary in $_BINDIR and if yes, check
     # if the version matches the expected one.
```
ggverma, line 14: `if yes, check if the version matches the expected one.`

Do we do this?

hack/get-kustomize.sh
```diff
@@ -5,21 +5,24 @@
...
-    local kustomize="$(PATH=$THIS_DIR command -v kustomize)"
+    local kustomize="$(PATH=$_BINDIR command -v kustomize)"
     if [ -x "$kustomize" ]; then
         >&2 echo "Found "$kustomize" version "`$kustomize version --short`
         echo $kustomize
         return 0
     fi
     local kustomize_url="https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz"
     curl -sLo kustomize.tar.gz "${kustomize_url}" || return 1
```
ggverma, line 22: `curl -sLo kustomize.tar.gz "${kustomize_url}" || return 1`
should this be
`curl -sLo $_BINDIR/kustomize.tar.gz "${kustomize_url}" || return 1`
and moved to line 24?

otherwise I think line 24 will fail. is my understanding right?

hack/get-kustomize.sh
salv-orlando: No, but we should because they have an habit of doing backward incompatible changes.

hack/get-kustomize.sh
salv-orlando: Line 21 downloads in the local directory, and line 24 expands it in $_BINDIR
But since I checked it locally but it seems I did not check it in my github profile I will verify.

salv-orlando: It appears to be working fine there too: https://github.com/salv-orlando/nsx-container-plugin-operator/runs/1938639109?check_suite_focus=true

hack/get-kustomize.sh
salv-orlando: PR updated

hack/get-kustomize.sh
ggverma: my bad. I though tar changes the directory before reading the tarball also https://www.gnu.org/software/tar/manual/html_node/Option-Summary.html#SEC42

ggverma (approved):

PR #106, opened 2021-02-17
Check and create NCP resources at startup

ggverma:
The user can delete nsx-ncp deployment or nsx-node-agent, nsx-ncp-bootstrap
Daemon Sets. If they are deleted when the operator is not running, the
operator should create them upon startup

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -332,26 +332,30 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 
 	if !needChange.ncp && !needChange.agent && !needChange.bootstrap {
 		// Check if NCP_IMAGE or nsx-ncp replicas changed
-		needChange.ncp, err = r.isNcpDeploymentChanged(ncpReplicas)
+		anyMonitoredResDeleted, err := r.isAnyMonitoredNCPResDeleted()
 		if err != nil {
 			return reconcile.Result{Requeue: true}, err
-		}
-
-		if !needChange.ncp {
-			log.Info("no new configuration needs to apply")
-			// Check if network config status must be updated
-			if networkConfigChanged {
-				err = updateNetworkStatus(networkConfig, instance, r)
-				if err != nil {
-					r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
-						fmt.Sprintf("Failed to update network status: %v", err))
-					return reconcile.Result{}, err
-				}
+		} else if !anyMonitoredResDeleted {
+			needChange.ncp, err = r.isNcpDeploymentChanged(ncpReplicas)
+			if err != nil {
+				return reconcile.Result{Requeue: true}, err
 			}
-			r.status.SetNotDegraded(statusmanager.ClusterConfig)
-			r.status.SetNotDegraded(statusmanager.OperatorConfig)
-			return reconcile.Result{}, nil
-		} else {
+
+			if !needChange.ncp {
+				log.Info("no new configuration needs to apply")
+				// Check if network config status must be updated
+				if networkConfigChanged {
+					err = updateNetworkStatus(networkConfig, instance, r)
+					if err != nil {
+						r.status.SetDegraded(statusmanager.ClusterConfig, "UpdateNetworkStatusError",
+							fmt.Sprintf("Failed to update network status: %v", err))
+						return reconcile.Result{}, err
+					}
+				}
+				r.status.SetNotDegraded(statusmanager.ClusterConfig)
+				r.status.SetNotDegraded(statusmanager.OperatorConfig)
+				return reconcile.Result{}, nil
+			}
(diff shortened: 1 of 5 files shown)


pkg/controller/configmap/configmap_controller.go
```diff
@@ -496,6 +500,25 @@ func (r *ReconcileConfigMap) updateSharedInfoWithNsxNcpResources(objs []*unstruc
 	log.Info("Updated shared info with Nsx Ncp Resources")
 }
 
+func (r *ReconcileConfigMap) isAnyPodMonitoredNCPResDeleted() (bool, error) {
```
salv-orlando, line 503: A very pedant comment: you can probably remove "Pod" from this func's name, since the function does not monitor pods after all?

pkg/controller/pod/pod_controller.go
```diff
@@ -221,53 +220,40 @@ func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, e
...
-		Name:      resName,
-	}
-
-	doesResExist, err := r.checkIfK8sResourceExists(instance, instanceDetails)
+	doesResExist, err := operatortypes.CheckIfNCPK8sResourceExists(
+		r.client, resName)
 	if err != nil {
 		log.Error(err, fmt.Sprintf(
-			"Could not retrieve K8s resource - '%s'", instanceDetails.Name))
+			"Could not retrieve K8s resource - '%s'", resName))
```
salv-orlando, line 227: another nit: this error message could be intented as "resource missing". I think we should rephrase to something like "an error occurred while retrieving K8s resource %s", resName

pkg/controller/pod/pod_controller.go
```diff
@@ -221,53 +220,40 @@ func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 		return err
 	}
 	if doesResExist {
 		log.Info(fmt.Sprintf(
-			"K8s resource - '%s' already exists", instanceDetails.Name))
+			"K8s resource - '%s' already exists", resName))
```
salv-orlando, line 232: more nits! This could go down to debug leve

pkg/controller/pod/pod_controller.go
```diff
@@ -221,53 +220,40 @@ func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 
-	log.Info(fmt.Sprintf("K8s resource - '%s' does not exist. It will be recreated", instanceDetails.Name))
+	log.Info(fmt.Sprintf("K8s resource - '%s' does not exist. It will be recreated", resName))
 
 	k8sObj := r.identifyAndGetK8SObjToCreate(resName)
 	if k8sObj == nil {
 		log.Info(fmt.Sprintf("%s spec not set. Waiting for config_map controller to set it", resName))
 	}
 	if err = r.setControllerReference(r, k8sObj); err != nil {
 		log.Info(fmt.Sprintf(
```
salv-orlando, line 243: sorry, I can't resist pedantry. since we are logging an error that would cause a failure in the recreation processes, let's log this at error level?

pkg/types/utils.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func identifyAndGetInstance(resName string) runtime.Object {
```
salv-orlando, line 36: opinion comment: we are encapsulating in this function the logic for leveraging the appropriate resource type.
This simplfies using CheckifNcpResourceExists, but also implies that if we should add any other resource here we will need to modify the func (and you have to find it first).

I suggest either:
1) keep type specification explicit as it was before
2) do the same but using types (NsxNcpBootstrapDsName could become a struct which also specifies it's a ds for instance (this might however require a bit of refactoring)
3) have explicit funcs "idenfiyandGetDeployment" and "identifyAndGetDaemonset"

but after all, this is just an opinion, so feel free to take it or ignore it!

pkg/controller/configmap/configmap_controller.go
ggverma: right, updated

pkg/controller/pod/pod_controller.go
ggverma: updated to `log.V(1)`

pkg/controller/pod/pod_controller.go
ggverma: agree ;)

pkg/types/utils.go
ggverma: I know, I thought about these but then it's unlikely we'll add a new resource and won't go over this path to include its logic. Still, we should future proof it. let me update

pkg/controller/pod/pod_controller.go
ggverma: done

salv-orlando (approved):

Issue #100, opened 2021-01-29
address verbose logging


salv-orlando:
This message is repeated about twice every second

```
{"level":"info","ts":"2021-01-29T14:04:52.149Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:52.279Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:52.524Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:52.858Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:53.858Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:53.946Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:54.289Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:54.551Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:54.710Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:54.869Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:55.000Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
{"level":"info","ts":"2021-01-29T14:04:55.945Z","logger":"controller_configmap","msg":"configmap controller watching events in namespace nsx-system-operator"}
```

timdengyun:
There are some points I'd like put:

[...]

PR #109, opened 2021-02-25
Add watchedNameSpace into controller Watch func to filer out unexpected namespace changes for ConfigMap Reconcile

timdengyun:
Issue brief: 
This verbose logging come from the Reconcile function of configmap_controller, which means there are many resources changes event watched by configmap controller in a short time,  as we see from issue#100,  twice every second or even worse in my testing. 

The **root cause** of this issue: 
There are many **unexpected** Configmap resource change events and Secretr resource change events from namespaces other than nsx-system-operator happened. These event changes also trigger also are pushed to handling queue of Reconcile process. That is why we see this verbose logging output from Reconcile function, since it was called multiple times to handle these events.  This issue usually happens during operator pod startup and configmap render phase. 

Solution of Identification:
By design, we added four types of resources change event to watch:
NcpInstall CRD
ConfigMap
Network CRD（openshift）
Secret

By design, we only want these resource change in configmap controller from a single namespace:
nsx-system-operator. 
However, from testing, ConfigMap type resource and Secret type resource also exist in the other namespace,
such as kube-system,  nsx-system, etc.

{"level":"info","ts":"2021-02-24T03:14:35.553Z","logger":"controller_configmap","msg":"Received unexpected namespace change","Request.Namespace":"kube-public","Request.Name":"cluster-info"}
{"level":"debug","ts":"2021-02-24T03:14:35.553Z","logger":"controller-runtime.controller","msg":"Successfully Reconciled","controller":"configmap-controller","request":"kube-public/cluster-info"}

cluster-info is Configmap type resource from kube-public namespace

{"level":"info","ts":"2021-02-24T10:29:39.819Z","logger":"controller_configmap","msg":"Received unexpected namespace change","Request.Namespace":"**kube-system**","Request.Name":"**service-controller-token-br68t**"}
{"level":"debug","ts":"2021-02-24T10:29:39.819Z","logger":"controller-runtime.controller","msg":"Successfully Reconciled","controller":"configmap-controller","request":"kube-system/service-controller-token-br68t"}

{"level":"info","ts":"2021-02-24T10:29:39.819Z","logger":"controller_configmap","msg":"Received unexpected namespace change","Request.Namespace":"kube-system","Request.Name":"coredns-token-fhg5g"}
{"level":"debug","ts":"2021-02-24T10:29:39.819Z","logger":"controller-runtime.controller","msg":"Successfully Reconciled","controller":"configmap-controller","request":"kube-system/coredns-token-fhg5g"}

{"level":"info","ts":"2021-02-24T10:29:39.819Z","logger":"controller_configmap","msg":"Received unexpected namespace change","Request.Namespace":"**nsx-system**","Request.Name":"**default-token-5ljpk**"}
{"level":"debug","ts":"2021-02-24T10:29:39.819Z","logger":"controller-runtime.controller","msg":"Successfully Reconciled","controller":"configmap-controller","request":"nsx-system/default-token-5ljpk"}

service-controller-token-br68t is Secrete type resource from kube-system namespace.
default-token-5ljpk is Secrete type resource from kube-system namespace.

Conclusion: We need to filter out unexpected resource changes from watched namespace:nsx-system-operator to avoid 
unnecessary Reconcile request, and then it will fix this verbose logging issue naturally. 

Solution:
Add a namespace filter for event changes for ConfigMap and Secret in Controller Watch function to  allow filter out the unexpected change event prior to enter event queue handler. 
This filter function will pass an our own implemented Predicate interface to Watch function as an argument. This will
check namespace of event coming from and only allow event from nsx-system-operator namespace to be passed.

In addition, the previous log output:
log.Info(fmt.Sprintf("configmap controller watching events in namespace %s", watchedNamespace))
is a little bit confused for users since we do have some events not from watchedNamespace:nsx-system-operator. 
for example, Network CRD. So I suggest that we move this log to the place where event request namespae is 
watchedNamespace.

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -27,9 +27,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -61,6 +63,28 @@ func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager, sha
 	return &reconcileConfigMap
 }
 
+// Create a Predicate with filter function used by Controllers to filter Events before they are provided to EventHandlers.
+// Only event objects matching the watchNameSpace will be selected, otherwise filtered out.
+func byNameSpaceFilter(watchNameSpace string) predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return (e.Meta.GetNamespace() == watchNameSpace)
+		},
+
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return (e.Meta.GetNamespace() == watchNameSpace)
+		},
+
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return (e.MetaNew.GetNamespace() == watchNameSpace)
+		},
+
+		GenericFunc: func(e event.GenericEvent) bool {
+			return (e.Meta.GetNamespace() == watchNameSpace)
+		},
+	}
+}
+
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Reconciler) error {
 	// Create a new controller
@@ -69,14 +93,29 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
 		return err
 	}
 
+	reconcileConfigMap, ok := r.(*ReconcileConfigMap)
+	if !ok {
+		log.Info("Reconciler Interface to ReconcileConfigMap failed")
+		return nil
+	}
+
+	watchedNamespace := reconcileConfigMap.status.OperatorNamespace
+	if watchedNamespace == "" {
+		// For NcpInstall, ConfigMap and Secret, we only want to watch a single namespace
+		// Pass watchedNamespace into controller Watch func to filer out unexpected namespace changes.
+		watchedNamespace = operatortypes.OperatorNamespace
+	}
+
 	// Watch for changes to primary resource NcpInstall CRD
-	err = c.Watch(&source.Kind{Type: &operatorv1.NcpInstall{}}, &handler.EnqueueRequestForObject{})
+	// The NcpInstall CRD changes may appear in mutiple namespaces, only changes from watched namespace of operator will be queued for handler
+	err = c.Watch(&source.Kind{Type: &operatorv1.NcpInstall{}}, &handler.EnqueueRequestForObject{}, byNameSpaceFilter(watchedNamespace))
 	if err != nil {
 		return err
 	}
 
 	// Watch for changes to primary resource ConfigMap
-	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{})
+	// The configmap resource changes may appear in mutiple namespaces, only changes from watched namespace of operator will be queued for handler
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, byNameSpaceFilter(watchedNamespace))
 	if err != nil {
 		return err
 	}
@@ -92,7 +131,8 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
 	}
 
 	// Watch for changes to primary resource Secret
-	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{})
+	// The Secret resouce changes may appear in mutiple namespaces, only changes from watched namespace of operator will be queued for handler
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, byNameSpaceFilter(watchedNamespace))
 	if err != nil {
 		return err
 	}
@@ -139,7 +179,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			operatortypes.OperatorNamespace))
 		watchedNamespace = operatortypes.OperatorNamespace
 	}
-	log.Info(fmt.Sprintf("configmap controller watching events in namespace %s", watchedNamespace))
+
 	if request.Namespace == watchedNamespace {
 		if request.Name == operatortypes.ConfigMapName {
 			reqLogger.Info("Reconciling nsx-ncp-operator ConfigMap change")
@@ -156,6 +196,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	} else if request.Namespace == "" && request.Name == operatortypes.NetworkCRDName {
 		reqLogger.Info("Reconciling cluster Network CRD change")
 	} else {
+		log.V(4).Info(fmt.Sprintf("Received change from unexpected namespace: %s", request.Namespace))
 		return reconcile.Result{}, nil
 	}
 


pkg/controller/configmap/configmap_controller.go
```diff
@@ -139,8 +165,10 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			operatortypes.OperatorNamespace))
 		watchedNamespace = operatortypes.OperatorNamespace
 	}
-	log.Info(fmt.Sprintf("configmap controller watching events in namespace %s", watchedNamespace))
```
timdengyun, line 142: It's better to move this line to place where request.Namespace == watchedNamespace

pkg/controller/configmap/configmap_controller.go
```diff
@@ -61,6 +63,28 @@ func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager, sha
 	return &reconcileConfigMap
 }
 
+// Creat a Predicate with filter function used by Controllers to filter Events before they are provided to EventHandlers. 
+// Only event objects matching the watachNameSpace will be selected, otherwise filtered out.
```
dantingl, line 67: watachNameSpace -> watchNamespace

pkg/controller/configmap/configmap_controller.go
```diff
@@ -76,7 +100,8 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
 	}
```
dantingl, line 100: Add byNameSpaceFilter for NcpInstall CRD?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -76,7 +100,8 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
...
 
 	// Watch for changes to primary resource ConfigMap
-	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{})
+	// The configmap resource changes may appear in mutiple namespaces, only changes from watched namespace of operator will be queued for handler
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, byNameSpaceFilter(operatortypes.OperatorNamespace))
```
dantingl, line 104: Below is the way to get watchedNamespace, do not use operatortypes.OperatorNamespace directly.
```
	watchedNamespace := r.status.OperatorNamespace
	if watchedNamespace == "" {
		log.Info(fmt.Sprintf("ConfigMapController can only watch a single namespace, defaulting to: %s",
			operatortypes.OperatorNamespace))
		watchedNamespace = operatortypes.OperatorNamespace
	}
```

pkg/controller/configmap/configmap_controller.go
```diff
@@ -139,8 +165,10 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			operatortypes.OperatorNamespace))
 		watchedNamespace = operatortypes.OperatorNamespace
 	}
-	log.Info(fmt.Sprintf("configmap controller watching events in namespace %s", watchedNamespace))
+	
 	if request.Namespace == watchedNamespace {
+		log.Info(fmt.Sprintf("Received configmap controller watching events in namespace: %s", watchedNamespace))
```
dantingl, line 170: I don't think we need this log as we have accurate log in each `if` block.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -156,6 +184,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	} else if request.Namespace == "" && request.Name == operatortypes.NetworkCRDName {
 		reqLogger.Info("Reconciling cluster Network CRD change")
 	} else {
+		log.Info(fmt.Sprintf("Received change from unexpected namespace: %s", request.Namespace))
```
dantingl, line 187: In which case operator will print this log?

heypnus:
Could you merge these 3 commits into one?

> bug fix for /issues/100 5096071
> fix typo in comments of byNameSpaceFilter 2c9623f
> change log output formate in Reconcile function a073a53

pkg/controller/configmap/configmap_controller.go
timdengyun: Thanks for correction.

pkg/controller/configmap/configmap_controller.go
timdengyun: So far, I don't find NcpInstall CRD type resources in other namespaces other than nsx-system-operator, however, it's a good defensive strategy to add byNameSpaceFilter for NcpInstall CRD as well, since it's supposed that we want to watch NcpInstall CRD only in nsx-system-operator as well.

pkg/controller/configmap/configmap_controller.go
timdengyun: Thanks a lot for advice.

pkg/controller/configmap/configmap_controller.go
timdengyun: I'm ok for  this one.

pkg/controller/configmap/configmap_controller.go
timdengyun: This is still a defensive log info. This log will be printed when there is unexpected event change from namespace other than "" or nsx-ncp-operator.  It should have no more such unexpected changes coming since we have added namespace filter for the controller. So the log info maybe incorrect, shall we put this as a debug level log?

timdengyun:
The latest file version already include these 3 commits? Is  there other method that I could merge these 3 commits? Thanks a lot.

> Could you merge these 3 commits into one?
> 
> > bug fix for [/issues/100](https://github.com/vmware/nsx-container-plugin-operator/issues/100) [5096071](https://github.com/vmware/nsx-container-plugin-operator/commit/509607135774f7d330d433a0fa1ab8acd7dd9a0a)
> > fix typo in comments of byNameSpaceFilter [2c9623f](https://github.com/vmware/nsx-container-plugin-operator/commit/2c9623f49e51114cf3465989e9a954dfaf61f485)
> > change log output formate in Reconcile function [a073a53](https://github.com/vmware/nsx-container-plugin-operator/commit/a073a53dbafa0a88e01ed8c69862f838af0f5cea)

timdengyun:
There are some points I'd like put: 

1. Using cache.MultiNamespacedCacheBuilder in options when creating Manager in main.go is not a correct way to 
 to manage MultiNamespae watch, this is better done via a Predicate. Moreover, we need to watch node change which is not namespace isolated and Network CRD type resource  in openshift. Therefore, we need a cluster-scoped watch operator rather than multiple namespace operator. So only options attribute: Namespace as ""  will be able to manage cluster-scoped CRs.

Improve opportunities: 
2. Controllers may Watch multiple Kinds of objects (e.g. Pods, ReplicaSets and Deployments), but they Reconciler only a single Type. Currently, our configmap controller Reconciler will watch four types of resources.  It's better to have a Reconciler for one resource type. that is to say, It is better to write Controllers to perform an O(1) Reconciler N times (e.g. on N different objects) instead of performing an O(N) Reconciler 1 time (e.g. on a single object which manages N other objects).
 
3.  There is a new func: NewPredicateFuncs in Predicate package from controller-runtime lib Version v0.6.2 onward, this is a simpler way compared to current fix approach. However, we are using  v0.5.2 that is missing this new func. we could 
refactor our code using this new one in the future.


4.  
4. Version v0.6.2

pkg/controller/configmap/configmap_controller.go
```diff
@@ -156,6 +195,7 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	} else if request.Namespace == "" && request.Name == operatortypes.NetworkCRDName {
 		reqLogger.Info("Reconciling cluster Network CRD change")
 	} else {
+		log.V(1).Info(fmt.Sprintf("Received change from unexpected namespace: %s", request.Namespace))
```
dantingl, line 198: how about setting log level to 4 or 6?

dantingl (commented):
Please remove previous 3 commits

pkg/controller/configmap/configmap_controller.go
timdengyun: Agreed. It's less important debug message.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -69,14 +93,29 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
 		return err
 	}
 
+	reconcileConfigMap, ok := r.(*ReconcileConfigMap)
+	if !ok {
+		log.Info("Reconciler Interface to ReconcileConfigMap pointer type cast assertion failed")
```
dantingl, line 98: Is there any case this could fail?

timdengyun: To my understanding, if passed argument:r is not a valid ReconcileConfigMap object, for example, nil object, or other struct type, which will lead to this type cast failed. I don't think there is other case causing fail.

timdengyun: This type cast assertion is compulsory.

heypnus (commented):
LGTM, just a minor comment.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -61,6 +63,28 @@ func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager, sha
 	return &reconcileConfigMap
 }
 
+// Creat a Predicate with filter function used by Controllers to filter Events before they are provided to EventHandlers.
```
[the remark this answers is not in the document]
timdengyun: Thanks a lot, fixed.

heypnus (approved):
LGTM, thanks

PR #111, opened 2021-03-02
[IPI] Disable coredns before local apiserver ready in master nodes

heypnus:
Currently, for OpenShift IPI, there will be a static pod keepalived
adding the Kubernetes VIP to the ovs_bridge on the master nodes and
the coredns pod running on every nodes to redirect the API request
to the VIP. However, the VIP works too early when the API server is
still only on bootstrap node and the port 6443 on master nodes are
not ready.

For the solution, there're 5 feasible manual workarounds:

1. Add the bootstrap IP to /etc/hosts on every master nodes then
   remove them when the apiserver on local becomes ready;
2. Deploy a forwarder on every master nodes to listen port 6443 and
   forward the traffic to bootstrap node;
3. Add an iptables rule on the port 53 of master nodes to temporarily
   make the coredns invalid then the apiserver can be resolved by
   external DNS server;
4. Change the order in /etc/resolv.conf to make the external DNS
   server at the top of local coredns, then the API server address
   will be resolved by external DNS server instead of local DNS.
5. Temperarily hack the static pod's yaml file for kubelet in master
   node to disable the keepalived, then restore them after local API
   server is ready.

To automate the above workarounds, for 1 and 2, it's complex to get the
accurate bootstrap IP; for 3 and 4, the DNS server should know the
bootstrap's IP but it knows only the VIP if the user is following
the official doc, and the /etc/resolv.conf is automatically generated by
NM so persisting it needs more effort. For 5, we need to change the
files in node, it's tricky. Taking into account the difficulties of user
configuration our changess, this patch is based on method 5. It will
create deamonset to deploy the pod on every master node then hack
the definiation of keepalived in /etc/kubernetes/manifests/keepalived.yaml
until loalhost:6443 is up.

The usage:
For the IPI customer, the deploy/openshift4/nsx-ipi-helper.yaml
should be additionally deployed. The customer not using IPI won't be
impacted.

### DIFF
--- a/deploy/openshift4/nsx-ipi-helper.yaml
+++ b/deploy/openshift4/nsx-ipi-helper.yaml
@@ -1,0 +1,58 @@
+# nsx-ipi-helper DaemonSet
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nsx-ipi-helper
+  namespace: nsx-system-operator
+spec:
+  updateStrategy:
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      name: nsx-ipi-helper
+  template:
+    metadata:
+      labels:
+        name: nsx-ipi-helper
+    spec:
+      hostNetwork: true
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+        - key: node.kubernetes.io/not-ready
+          effect: NoSchedule
+        - key: node.kubernetes.io/unreachable
+          effect: NoSchedule
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      volumes:
+        - name: host-manifests
+          hostPath:
+            path: "/etc/kubernetes/manifests"
+      initContainers:
+        - name: nsx-ipi-helper
+          image: registry.access.redhat.com/ubi8/ubi:latest
+          imagePullPolicy: IfNotPresent
+          command:
+          - bin/bash
+          - -c
+          - |
+            while true; do
+              sed -i "s/-f \/etc\/keepalived\/keepalived.conf/-f \/etc\/keepalived\/fake_keepalived.conf/" /etc/kubernetes/manifests/keepalived.yaml && echo `date` Temporarily disabled keepalived
+              if /usr/bin/curl -kLfs https://localhost:6443/readyz; then
+                sed -i "s/fake_keepalived.conf/keepalived.conf/" /etc/kubernetes/manifests/keepalived.yaml
+                echo `date` Restored the keepalived in localhost
+                exit 0
+              fi
+              sleep 1
+            done
+          volumeMounts:
+            - name: host-manifests
+              mountPath: "/etc/kubernetes/manifests"
+          securityContext:
+            privileged: true
+      containers:
+        - name: nsx-ipi-dummy
+          image: registry.access.redhat.com/ubi8/ubi:latest
+          imagePullPolicy: IfNotPresent
+          command: ["/bin/bash", "-c", "while true; do sleep 5; done"]
--- a/README.md
+++ b/README.md
@@ -61,9 +61,10 @@ Create manifest files:
 ```
 $ openshift-install --dir=MY_CLUSTER create manifests
 ```
-Put operator yaml files from `deploy/openshift4/` to `MY_CLUSTER/manifests`,
+Copy operator yaml files from `deploy/openshift4/` to `MY_CLUSTER/manifests`,
 edit configmap.yaml about operator configurations, add the operator image and
-NCP image in operator.yaml.
+NCP image in operator.yaml. Note `deploy/openshift4/nsx-ipi-helper.yaml` is only
+for IPI, if you are using UPI please don't copy it.
 
 Generate ignition configuration files:
 ```


deploy/openshift4/nsx-ipi-helper.yaml
```diff
@@ -0,0 +1,51 @@
...
+          effect: NoSchedule
+        - key: node.kubernetes.io/not-ready
+          effect: NoSchedule
+        - key: node.kubernetes.io/unreachable
+          effect: NoSchedule
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      initContainers:
+        - name: nsx-ipi-helper
+          image: nsx-ujo-docker-local.artifactory.eng.vmware.com/3.1.1.17415517/nsx-ncp-rhel:latest
```
heypnus, line 38: It's a temp image for verification.

heypnus: Didn't find an offical image including both curl and iptables so I'm using ubi8 and exec "yum -y install iptables" in the command...

deploy/openshift4/nsx-ipi-helper.yaml
```diff
@@ -0,0 +1,49 @@
...
+          effect: NoSchedule
+        - key: node.kubernetes.io/unreachable
+          effect: NoSchedule
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      initContainers:
+        - name: nsx-ipi-helper
+          image: registry.access.redhat.com/ubi8/ubi:latest
+          imagePullPolicy: IfNotPresent
+          command: ["/bin/bash", "-c", "yum -y install iptables; echo Disabling the DNS in localhost; iptables -A INPUT -p udp --dport 53 -j DROP; iptables -S INPUT; while ! curl localhost:6443 ; do sleep 1 ; done; echo Enabling the DNS in localhost; iptables -D INPUT -p udp --dport 53 -j DROP; iptables -S INPUT; echo Done"]
```
ggverma, line 40: Is only UDP 53 need to be deactivated/activated? What about TCP? Also, not sure if port 9153 should also be added here (likely not)

ggverma, line 40: consider installing it in the base image. Installing packages at runtime adds the risk of incompatibility, dependency, and other unknown issues

ggverma, line 40: I think using `&&` instead of `;` to execute instructions will be better since we are not using a script and we do not fall into unknown erroneous state. Alternatively, you can convert this into a small script

ggverma (commented):
Perhaps we can also delete this DS using operator when nsx-ipi-dummy comes into Running status?

ggverma:
Or maybe we shouldn't delete it if it's needed during Upgrade..?

deploy/openshift4/nsx-ipi-helper.yaml
```diff
@@ -0,0 +1,49 @@
+# nsx-ipi-helper DaemonSet
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nsx-ipi-helper
+  namespace: nsx-system-operator
+  labels:
+    tier: nsx-networking
```
dantingl, line 8: No need to add `nsx-networking` and `v1` lebal, or maybe we can just remove `labels`?

deploy/openshift4/nsx-ipi-helper.yaml
dantingl: If installing package at runtime is not a good approach, I think maybe we can directly use NCP image if we cannot find a good upstream image with ubi+iptables. I don't think it's worth to maintain a new image for IPI support.

dantingl: Adding a script means we need to maintain a Dockerfile for IPI installer support.

heypnus: In my verification only deactivating port 53 could work (9153 is not in listening in a running cluster, not sure if it was up during bootstrap), e.g.

> [root@cluster46-q5ntm-worker-hs5sc ~]# **dig api.cluster46.openshift.test**
> 
> ; <<>> DiG 9.11.13-RedHat-9.11.13-6.el8_2.1 <<>> api.cluster46.openshift.test
> ;; global options: +cmd
> ;; Got answer:
> ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 11746
> ;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
> ;; WARNING: recursion requested but not available
> 
> ;; OPT PSEUDOSECTION:
> ; EDNS: version: 0, flags:; udp: 4096
> ; COOKIE: ab76f2aec2f0ad74 (echoed)
> ;; QUESTION SECTION:
> ;api.cluster46.openshift.test.  IN      A
> 
> ;; ANSWER SECTION:
> api.cluster46.openshift.test. 14 IN     A       192.169.41.11
> 
> ;; Query time: 11 msec
> ;; SERVER: 192.169.41.101#53(192.169.41.101)
> ;; WHEN: Thu Mar 04 06:47:11 UTC 2021
> ;; MSG SIZE  rcvd: 113
> 
> [root@cluster46-q5ntm-worker-hs5sc ~]# **iptables -A INPUT -p udp --dport 53 -j DROP**
> [root@cluster46-q5ntm-worker-hs5sc ~]# **dig api.cluster46.openshift.test**                                                                                                                           
> 
> ; <<>> DiG 9.11.13-RedHat-9.11.13-6.el8_2.1 <<>> api.cluster46.openshift.test
> ;; global options: +cmd
> ;; Got answer:
> ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 6058
> ;; flags: qr aa rd ra; QUERY: 1, ANSWER: 10, AUTHORITY: 2, ADDITIONAL: 4
> 
> ;; OPT PSEUDOSECTION:
> ; EDNS: version: 0, flags:; udp: 4096
> ; COOKIE: 61224b19821debf5901fffa9604082955c325f251bd61990 (good)
> ;; QUESTION SECTION:
> ;api.cluster46.openshift.test.  IN      A
> 
> ;; ANSWER SECTION:
> api.cluster46.openshift.test. 604800 IN A       192.169.41.105
> api.cluster46.openshift.test. 604800 IN A       192.169.41.108
> api.cluster46.openshift.test. 604800 IN A       192.169.41.103
> api.cluster46.openshift.test. 604800 IN A       192.169.41.106
> api.cluster46.openshift.test. 604800 IN A       192.169.41.101
> api.cluster46.openshift.test. 604800 IN A       192.169.41.104
> api.cluster46.openshift.test. 604800 IN A       192.169.41.102
> api.cluster46.openshift.test. 604800 IN A       192.169.41.107
> api.cluster46.openshift.test. 604800 IN A       192.169.41.109
> api.cluster46.openshift.test. 604800 IN A       192.169.41.110
> 
> ;; AUTHORITY SECTION:
> openshift.test.         604800  IN      NS      dns1.openshift.test.
> openshift.test.         604800  IN      NS      localhost.
> 
> ;; ADDITIONAL SECTION:
> dns1.openshift.test.    604800  IN      A       10.92.204.129
> localhost.              604800  IN      A       127.0.0.1
> localhost.              604800  IN      AAAA    ::1
> 
> ;; Query time: 1 msec
> ;; SERVER: 10.40.121.149#53(10.40.121.149)
> ;; WHEN: Thu Mar 04 06:47:46 UTC 2021
> ;; MSG SIZE  rcvd: 347

heypnus: I also don't want to add a new image, NCP image also could work but if we use it, it seems that we need the customer to set this value manually? I'm not sure what's the best:
1. create a new customized image for IPI
2. use the official UBI image then install the package
3. use NCP image but need the user to set the image manually

heypnus: `&&` looks better I'll try it.

deploy/openshift4/nsx-ipi-helper.yaml
heypnus: OK, I just copy it from nsx-node-agent, in fact I also think it's not useful... will remove the labels.

deploy/openshift4/nsx-ipi-helper.yaml
heypnus: For now, the patch is using 2. @salv-orlando Do you have any comment?

ggverma: can we use operator to create this DS?

heypnus:
Abandoning this PR because after the deploying there's still VIP conflict issue.

PR #112, opened 2021-03-18
Improve the resync logic for node network status

heypnus:
Sometimes when a nsx-node-agent pod is created and
running for less than 180 seconds, the operator will
try to update the node status twice (firstly set
network-unavailable=true, then sleep and try to set
network-unavailable=false after 180 seconds)[1].

The code has a redundant check before sleeping, and
in the check logic, Get API reads the node status
from cache which may be not synced after the first
update operation was executed, so an unexpected
"Node condition is not changed" will be reported,
then the taints cannot be removed until the removal
logic was triggerred accidentally by another event
from nsx-node-agent pod. This patch will remove the
redundant check.

And we will assume that the data read by client will
eventually be correct, but may be slightly out of
date. So this patch introduced the logic
assertNodeStatus to ensure the final status is
expected.

This patch also replace the goroutine with RequeueAfter,
the latter is a more native and less error-prone
implementation.

[1] The following logs show this case:

{"level":"info","ts":"2021-03-08T14:56:37.864Z","logger":"status_manager","msg":"nsx-node-agent-p8ss5/nsx-kube-proxy for node compute-2 started for less than 17.864554094s"}
{"level":"info","ts":"2021-03-08T14:56:37.864Z","logger":"status_manager","msg":"nsx-node-agent-p8ss5/nsx-node-agent for node compute-2 started for less than 17.864554094s"}
{"level":"info","ts":"2021-03-08T14:56:37.864Z","logger":"status_manager","msg":"nsx-node-agent-p8ss5/nsx-ovs for node compute-2 started for less than 17.864554094s"}
{"level":"info","ts":"2021-03-08T14:56:37.864Z","logger":"status_manager","msg":"Setting status NetworkUnavailable to true for node compute-2"}
{"level":"info","ts":"2021-03-08T14:56:37.876Z","logger":"status_manager","msg":"Updated node condition NetworkUnavailable to true for node compute-2"}
{"level":"info","ts":"2021-03-08T14:56:37.876Z","logger":"status_manager","msg":"Node condition is not changed"}
...
{"level":"info","ts":"2021-03-08T15:26:13.541Z","logger":"status_manager","msg":"Setting status NetworkUnavailable to false for node compute-2"}
{"level":"info","ts":"2021-03-08T15:26:13.541Z","logger":"status_manager","msg":"Setting status NetworkUnavailable to false for node compute-2 after -26m53.541741583s"}

### DIFF
--- a/pkg/controller/pod/pod_controller.go
+++ b/pkg/controller/pod/pod_controller.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"strings"
-	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/apply"
@@ -41,11 +40,6 @@ var log = logf.Log.WithName("controller_pod")
 var SetControllerReference = controllerutil.SetControllerReference
 
 var ApplyObject = apply.ApplyObject
-
-// The periodic resync interval.
-// We will re-run the reconciliation logic, even if the NCP configuration
-// hasn't changed.
-var ResyncPeriod = 2 * time.Minute
 
 var firstBoot = true
 
@@ -189,15 +183,17 @@ func (r *ReconcilePod) isForNsxNodeAgentPod(request reconcile.Request) bool {
 func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
-	if err := r.status.CheckExistingAgentPods(&firstBoot, r.sharedInfo); err != nil {
-		return reconcile.Result{Requeue: true}, err
+	result, err := r.status.CheckExistingAgentPods(&firstBoot, r.sharedInfo)
+	emptyResult := reconcile.Result{}
+	if result != emptyResult || err != nil {
+		return result, err
 	}
 
 	if !r.isForNcpDeployOrNodeAgentDS(request) {
 		// the request is not for ncp deployement or nsx-node-agent ds, but for nsx-node-agent pod
 		if r.isForNsxNodeAgentPod(request) {
 			reqLogger.Info("Reconciling pod update for network status")
-			r.status.SetNodeConditionFromPod(request.NamespacedName, r.sharedInfo, nil)
+			return r.status.SetNodeConditionFromPod(request.NamespacedName, r.sharedInfo, nil)
 		}
 		return reconcile.Result{}, nil
 	}
@@ -216,7 +212,7 @@ func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, e
 		}
 	}
 
-	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
+	return reconcile.Result{RequeueAfter: operatortypes.DefaultResyncPeriod}, nil
 }
 
 func (r *ReconcilePod) recreateNsxNcpResourceIfDeleted(resName string) error {
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	client "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 var log = logf.Log.WithName("status_manager")
@@ -64,7 +65,7 @@ type deploymentState struct {
 	LastChangeTime time.Time
 }
 
-func FindStatusCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
+func FindNodeStatusCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
 	for i := range conditions {
 		if conditions[i].Type == conditionType {
 			return &conditions[i]
@@ -78,7 +79,7 @@ func SetNodeCondition(conditions *[]corev1.NodeCondition, newCondition corev1.No
 	if conditions == nil {
 		conditions = &[]corev1.NodeCondition{}
 	}
-	existingCondition := FindStatusCondition(*conditions, newCondition.Type)
+	existingCondition := FindNodeStatusCondition(*conditions, newCondition.Type)
 	if existingCondition == nil {
 		newCondition.LastTransitionTime = metav1.NewTime(time.Now())
 		*conditions = append(*conditions, newCondition)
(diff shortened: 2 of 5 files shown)


pkg/controller/pod/pod_controller.go
```diff
@@ -179,6 +179,13 @@ func (r *ReconcilePod) isForNsxNodeAgentPod(request reconcile.Request) bool {
 	if request.Namespace == operatortypes.NsxNamespace && strings.Contains(
 		request.Name, operatortypes.NsxNodeAgentDsName) &&
 		request.Name != operatortypes.NsxNodeAgentDsName {
+		// Check if the pod exists to prevent that a stale pod is in a endless periodic resync
```
dantingl, line 182: If nsx-node-agent pod is deleted, could you set node status to False in this case?

pkg/controller/pod/pod_controller.go
```diff
@@ -198,6 +205,7 @@ func (r *ReconcilePod) Reconcile(request reconcile.Request) (reconcile.Result, e
 		if r.isForNsxNodeAgentPod(request) {
 			reqLogger.Info("Reconciling pod update for network status")
 			r.status.SetNodeConditionFromPod(request.NamespacedName, r.sharedInfo, nil)
+			return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
```
dantingl, line 208: If you fix the issue to get node status after operator updates it, do we still need to reconcile in a loop?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -144,12 +144,7 @@ func (status *StatusManager) setNodeNetworkUnavailable(nodeName string, ready bo
 	} else {
 		delete(sharedInfo.LastNetworkAvailable, nodeName)
 	}
-	_, changed := status.combineNode(nodeName, ready, reason, message)
-	if !changed {
-		log.Info("Node condition is not changed")
-		return
-	}
-	log.Info(fmt.Sprintf("Setting status NetworkUnavailable to %v for node %s", !ready, nodeName))
+	log.V(1).Info(fmt.Sprintf("Setting status NetworkUnavailable to %v for node %s", !ready, nodeName))
```
dantingl, line 147: Please file an issue to update all logs with log level.

pkg/controller/pod/pod_controller.go
heypnus: When the operator checks network-unavailable status, it won't update the status if it fails to get the nsx-node-agent pod because this case is usually that the nsx-node-agent is updated, i.e. old pod is deleted and new pod is created, it's safe to ignore the event from old pod.

pkg/controller/pod/pod_controller.go
heypnus: After the fix, the logic will be:
1. the nsx-node-agent pod is running for less than 180s, temporarily update the network-unavailable to true
2. sleep until 180s
3. finally update the network-unavailable to false
-- in step 2, a common case is that the time is slightly less than 180s, e.g. sleep 160s, but a corner case may be sleep 0.001s. For the corner case, I think we should have a requeue mechanism to double secure it.

In fact, I think the ideal implementation is that we don't invoke more than one update operation in one reconcile, but that needs to refactor for the whole logic,  we cannot complete it in this release.

heypnus: Synced offline and uploaded a new patch set which refactored the logic, replacing goroutine with ResyncAfter

pkg/controller/statusmanager/pod_status.go
```diff
@@ -98,77 +99,96 @@ func SetNodeCondition(conditions *[]corev1.NodeCondition, newCondition corev1.No
 // CheckExistingAgentPods is for a case: nsx-node-agent becomes unhealthy -> NetworkUnavailable=True -> operator off ->
 // nsx-node-agent becomes healthy and keeps running -> operator up -> operator cannot receive nsx-node-agent event to
 // set NetworkUnavailable=False. So a full sync at the start time is necessary.
-func (status *StatusManager) CheckExistingAgentPods(firstBoot *bool, sharedInfo *sharedinfo.SharedInfo) error {
+func (status *StatusManager) CheckExistingAgentPods (firstBoot *bool, sharedInfo *sharedinfo.SharedInfo) (reconcile.Result, error) {
```
dantingl, line 102: `CheckExistingAgentPods (` has a whitespace

pkg/controller/statusmanager/pod_status.go
```diff
@@ -98,77 +99,96 @@ func SetNodeCondition(conditions *[]corev1.NodeCondition, newCondition corev1.No
...
 	}
 	*firstBoot = false
-	return nil
+	return reconcile.Result{}, nil
 }
 
-func (status *StatusManager) setNodeNetworkUnavailable(nodeName string, ready bool, startedAt *time.Time, reason string, message string, sharedInfo *sharedinfo.SharedInfo) {
-	if sharedInfo.LastNetworkAvailable == nil {
-		sharedInfo.LastNetworkAvailable = make(map[string]time.Time)
+// In general, write your controller with the assumption that information will eventually be correct, but may be slightly out of date.
```
dantingl, line 144: Please start with function name for comments

pkg/controller/pod/pod_controller.go
heypnus: reverted this change in new patch set

pkg/controller/statusmanager/pod_status.go
```diff
@@ -98,77 +99,96 @@ func SetNodeCondition(conditions *[]corev1.NodeCondition, newCondition corev1.No
...
-		status.updateNodeStatus(nodeName, ready, reason, message)
-		return
-	}()
+		return status.assertNodeStatus(nodeName, ready)
+	}
+	if err := status.updateNodeStatus(nodeName, false, reason, message); err != nil {
+		return reconcile.Result{}, err
+	}
+	sleepTime := operatortypes.TimeBeforeRecoverNetwork - startedTime
+	log.Info(fmt.Sprintf("Waiting %v to double check network status for node %s", sleepTime, nodeName))
```
dantingl, line 190: How about using debug level for this log

pkg/controller/statusmanager/pod_status.go
heypnus: https://github.com/vmware/nsx-container-plugin-operator/issues/117 was filed

dantingl (approved):

pkg/controller/statusmanager/pod_status.go
heypnus: Done

pkg/controller/statusmanager/pod_status.go
heypnus: Done

pkg/controller/statusmanager/pod_status.go
heypnus: Done

PR #113, opened 2021-03-23
Expand the matching range for searching node LSP

heypnus:
Previously we only matched the node address via ExternalIP.
This patch will expand the matching range to all node
addresses, including ExternalIP, InternalIP, etc.

Appending some examples of node addresses info:
1. [{ExternalIP 192.168.10.38} {InternalIP 192.168.10.38} {Hostname compute-0}]}
2. [{InternalIP 172.10.0.3} {Hostname vm10-dhcp-0-127}]}

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -6,6 +6,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"crypto/tls"
@@ -114,7 +115,7 @@ var cachedNodeSet = map[string](*statusmanager.NodeStatus){}
 func getNodeExternalIdByProviderId(nsxClients *NsxClients, nodeName string, providerId string) (string, error) {
 	// providerId has the following format: vsphere://<uuid>
 	if len(providerId) != 46 {
-		return "", errors.Errorf("Invalid provider ID %s of node %s", providerId, nodeName)
+		return "", errors.Errorf("invalid provider ID %s of node %s", providerId, nodeName)
 	}
 	providerId = string([]byte(providerId)[10:])
 	nsxClient := nsxClients.ManagerClient
@@ -130,7 +131,7 @@ func getNodeExternalIdByProviderId(nsxClients *NsxClients, nodeName string, prov
 			}
 		}
 	}
-	return "", errors.Errorf("No virtual machine matches provider ID %s and hostname %s", providerId, nodeName)
+	return "", errors.Errorf("no virtual machine matches provider ID %s and hostname %s", providerId, nodeName)
 }
 
 func listAttachmentsByNodeExternalId(nsxClients *NsxClients, vmExternalId string) ([]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -531,73 +545,73 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
-	foundClusterTag := false
 	nodeNameScope := "ncp/node_name"
 	clusterScope := "ncp/cluster"
-	for _, tag := range lsp.Tags {
-		if tag.Scope == nodeNameScope && tag.Tag == request.Name {
-			foundNodeTag = true
-		} else if tag.Scope == clusterScope && tag.Tag == cluster {
-			foundClusterTag = true
+	anyUpdate := false
+	for _, lsp := range(lsps) {
```
dantingl, line 568: Does it work for NCP if you tag multiple LSP with the same node name?

pkg/controller/node/node_controller.go
```diff
@@ -531,73 +545,73 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
-			Address: nodeAddress,
+			Addresses: nodeAddresses,
 			Success: false,
 			Reason:  fmt.Sprintf("Error while achieving ports for node %s: %v", nodeName, err),
 		}
 		r.status.SetFromNodes(cachedNodeSet)
 		return reconcile.Result{}, err
 	}
-	lsp, err := filterPortByNodeAddress(nsxClients, portList, nodeAddress)
+	lsps, err := filterPortsByNodeAddresses(nsxClients, portList, nodeAddresses)
```
dantingl, line 555: Even node has many addresses including external IP and internal IP, I don't think we need get addresses from multiple LSPs. It should be simpler if we have the limitation that node only attaches one vNIC to LS.

pkg/controller/node/node_controller.go
heypnus: Synced offline and the new patch set added the check

pkg/controller/node/node_controller.go
heypnus: ditto

pkg/controller/node/node_controller.go
```diff
@@ -177,26 +178,46 @@ func listPortsByAttachmentIds(nsxClients *NsxClients, attachmentIds []string) (*
...
 		}
 	}
-	return nil, errors.Errorf("No port matches address %s", nodeAddress)
-
+	var err error
+	lspCount := len(filteredPorts)
+	if lspCount == 0 {
+		err = errors.Errorf("No port matches addresses %v", nodeAddresses)
+	} else if lspCount > 1 {
+		err = errors.Errorf("Error while search logical port for addresses %v, expecting 1, got %d", nodeAddresses, lspCount)
```
dantingl, line 218: lower case for error message

pkg/controller/node/node_controller.go
```diff
@@ -477,19 +498,15 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
-	// TODO: An example for the nodeAddresses is:
-	// [{ExternalIP 192.168.10.38} {InternalIP 192.168.10.38} {Hostname compute-0}]"}
-	// Need to confirm if it's possible to have multiple ExternalIPs and handle that case
-	reqLogger.Info("Got the node address info", "nodeAddresses", nodeAddresses)
-	for _, address := range nodeAddresses {
-		if address.Type == "ExternalIP" {
-			nodeAddress = address.Address
-			break
+	nodeAddressesWithType := instance.Status.Addresses
+	var nodeAddresses []string
```
dantingl, line 502: Right now I'm thinking as we use node IP address to get node's LSP, it has a limitation that k8s cluster is using nsx-attached vNIC. In this case I guess there is only one IP for the node, either in ExternalIP or InternalIP? But using `nodeAddresses` as a list here should be no harm.

dantingl (approved):

pkg/controller/node/node_controller.go
heypnus: done

pkg/controller/node/node_controller.go
heypnus: Yes, currently we only have one IP. I use the list because of the possible expansion in the future.

dantingl (approved):

PR #116, opened 2021-03-24
Change secret compare func name

timdengyun:
Correct secret compare function name in consistent with function intention:

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -588,7 +588,7 @@ func (r *ReconcileConfigMap) isSecretChanged(opNsxSecret *corev1.Secret, opLbSec
 		}
 		ncpNsxSecret = nil
 	}
-	nsxSecretChanged := secretEqual(opNsxSecret, ncpNsxSecret)
+	nsxSecretChanged := secretNotEqual(opNsxSecret, ncpNsxSecret)
 	if nsxSecretChanged {
 		log.Info("nsx-secret is changed")
 		return true, nil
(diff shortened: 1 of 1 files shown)


pkg/controller/configmap/configmap_controller.go
```diff
@@ -588,7 +588,7 @@ func (r *ReconcileConfigMap) isSecretChanged(opNsxSecret *corev1.Secret, opLbSec
 		}
 		ncpNsxSecret = nil
 	}
-	nsxSecretChanged := secretEqual(opNsxSecret, ncpNsxSecret)
+	nsxSecretChanged := secretNotEqual(opNsxSecret, ncpNsxSecret)
```
timdengyun, line 591: I believe func name :secretEqual is not semantically correct according to function purpose.
I suggest it be changed to secretNotEqual although its intrinsic logic is correct.

dantingl: How about `isSecretChanged`?

timdengyun: We already have a func named by:isSecretChanged
func (r *ReconcileConfigMap) isSecretChanged(opNsxSecret *corev1.Secret, opLbSecret *corev1.Secret) (bool, error) {
...

secretNotEqual func is called in this func. Moreover, this func is used to compare secret content. 

Otherwise, we change logic of secretEqual to make it return true when equal, it make more intuitive. 
In such way,

nsxSecretEqual := secretEqual(opNsxSecret, ncpNsxSecret)
if !nsxSecretEqual {
 		log.Info("nsx-secret is changed")
 		return true, nil
     .........

heypnus (approved):

PR #119, opened 2021-03-30
Immediately respond when the customer correct error in role

heypnus:
The previous error handling logic is that when the operator hit
error during applying configuration, the reconciler will receive
an error then requeue the request. However, the requeue mechanism
has a potential backoff, that caused the operator may not
immediately respond when the customer correct error in operator's
cluster role.

This patch will additional watch the cluster role change for
the immediate response.

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -20,6 +20,7 @@ import (
 	operatortypes "github.com/vmware/nsx-container-plugin-operator/pkg/types"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


pkg/controller/configmap/configmap_controller.go
```diff
@@ -189,6 +189,10 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 			reqLogger.Info("Reconciling nsx-secret change")
 		} else if request.Name == operatortypes.LbSecret {
 			reqLogger.Info("Reconciling lb-secret change")
+		} else if request.Name == operatortypes.OperatorRoleName {
+			// The operator's ClusterRole/ClusterRoleBinding/ServiceAccount have the same name "nsx-ncp-operator",
```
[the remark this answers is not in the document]
heypnus: Insightful comment. Fixed, thanks!

pkg/controller/configmap/configmap_controller.go
```diff
@@ -137,6 +138,12 @@ func add(mgr manager.Manager, sharedInfo *sharedinfo.SharedInfo, r reconcile.Rec
 		return err
 	}
 
+	// Watch for changes to primary resource ClusterRole
+	err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestForObject{})
```
dantingl, line 142: Adding `byNameSpaceFilter(watchedNamespace)` as other watchers?

heypnus: clusterrole is a cluster level object, we should not add the namespace filter for it, otherwise, the configmap reconciler cannot receive the clusterrole event.

PR #120, opened 2021-03-31
Improvements for SecretNotEqual function

salv-orlando:
Improve logic for comparing keys in the secret in order to avoid
conditions where empty and missing tls keys (eg.: tls.ca) are interpreted
as differences in the secret.

Also, do not use DeepEqual as there is only need for comparing tls data
in the secret. This will avoid unnecessarily restarting ncp for unrelated
operator secret changes.

Finally do not include tls.ca in the projected volume if not specified in
the operator secret, or empty.

This commit also adds unit tests for secret comparison cases not
previously covered

### DIFF
--- a/pkg/controller/configmap/configmap_controller_test.go
+++ b/pkg/controller/configmap/configmap_controller_test.go
@@ -132,13 +132,24 @@ func TestConfigMapController_isSecretChanged(t *testing.T) {
 	secretChanged, _ = r.isSecretChanged(nsxSecret, nil)
 	assert.True(t, secretChanged)
 
-	// Secret equal case
+	// Secret equal case, with missing key
 	secretChanged, _ = r.isSecretChanged(nsxSecret, lbSecret)
 	assert.False(t, secretChanged)
-
-	// Secret not equal case
+	// Secret equal, with empty key
 	mockSecret := &corev1.Secret{
+		Data: map[string][]byte{"tls.crt": mockValue, "tls.key": mockValue, "tls.ca": []byte{}},
+	}
+	secretChanged, _ = r.isSecretChanged(nsxSecret, mockSecret)
+	assert.False(t, secretChanged)
+	// Secret not equal case, with missing key
+	mockSecret = &corev1.Secret{
 		Data: map[string][]byte{"tls.crt": mockValue, "tls.key": []byte("key")},
+	}
+	secretChanged, _ = r.isSecretChanged(nsxSecret, mockSecret)
+	assert.True(t, secretChanged)
+	// Secret not equal, with all keys
+	mockSecret = &corev1.Secret{
+		Data: map[string][]byte{"tls.crt": mockValue, "tls.key": []byte("key"), "tls.ca": mockValue},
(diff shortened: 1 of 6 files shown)


ggverma (approved):

pkg/controller/configmap/configmap_controller_test.go
```diff
@@ -132,16 +132,27 @@ func TestConfigMapController_isSecretChanged(t *testing.T) {
...
+	// Secret equal case, with missing key
 	secretChanged, _ = r.isSecretChanged(nsxSecret, lbSecret)
 	assert.False(t, secretChanged)
-
-	// Secret not equal case
+	// Secret equal, with empty key
 	mockSecret := &corev1.Secret{
+		Data: map[string][]byte{"tls.crt": mockValue, "tls.key": mockValue, "tls.ca": []byte{}},
+	}
+	secretChanged, _ = r.isSecretChanged(nsxSecret, mockSecret)
```
heypnus, line 142: In this case, nsxSecret.Data is `{"tls.crt": mockValue, "tls.key": mockValue, "tls.ca": mockValue}`, mockSecret.Data is `{"tls.crt": mockValue, "tls.key": mockValue, "tls.ca": []byte{}}`, looks that `nsxSecret.Data["tls.ca"]` is not equal to `mockSecret.Data["tls.ca"]` so `the secretChanged` should be True? I'm confused why `assert.False(t, secretChanged)` in line 143 passed... I checked the logic and didn't find any problem, seems that I missed something...

salv-orlando: That's exactly the main thing I changed with this patch.

Basically the root cause of the problem is that since the user was not specifying tls.ca at all, the render was creating an empty entry for tls.ca, and the comparison would always return False.

I made a patch for changing the renderer to reflect exactly what the user specified, but the resulting change was a bit ugly. The reason is that the golang text template don't distinguish well between nil and empty. So I had to introduce a "special" value meaning "missing" key and add somewhat complex logic in the template.

I think it's much easier to perform secret equality checks in this way: focusing only on the three keys we use and considering an empty key tantamount to a missing key

heypnus: Hi Salvatore, I understand your implementation about comparing keys one by one, but my question is more specific:
In line 142, the code is comparing `nsxSecret` and `mockSecret`. From the previous context, I think the real value of `nsxSecret.Data` is `{"tls.crt": ..., "tls.key": ..., "tls.ca": []byte("mockCrt")}` (according to line line 105 and 112), the real value of `mockSecret.Data` is `{"tls.crt": ..., "tls.key": ..., "tls.ca": []byte{}}` (according to line line 140). What confused me is that why `[]byte("mockCrt")` is equal to `[]byte{}`

heypnus: Ignore this comment. I dived into the code and found it's exactly comparing the `lbSecret` and `mockSecret`, not `nsxSecret` and `mockSecret`...

PR #123, opened 2021-04-02
Update operator configmap from the automatically generated yaml

heypnus:


deploy/kubernetes/configmap.yaml
```diff
@@ -6,31 +6,79 @@ metadata:
...
 
-    # VC API server HTTPS port.
-    #https_port = 443
 
 
     [coe]
 
     # Container orchestrator adaptor to plug in.
-    adaptor = kubernetes
+    #adaptor = kubernetes
```
dantingl, line 71: Why comment `adaptor`

deploy/kubernetes/configmap.yaml
```diff
@@ -6,31 +6,79 @@ metadata:
...
 
     # Specify cluster for adaptor.
-    cluster = k8scl-one
+    #cluster = k8scluster
```
dantingl, line 74: It's better to set `k8scluster` as default value. But do you think if this change will have upgrade regression concern?

deploy/kubernetes/configmap.yaml
heypnus: Because it's automatically generated from the jinja2 template in NCP. i.e. in the other all-in-one yaml, there's also `#adaptor = kubernetes`. And in NCP config we had defined the default value to "kubernetes" so it won't introduce issue. I don't want to uncomment it because if we do it we should uncomment it every time when updating the file, unless we change the jinja2 template in NCP, which change also will impact the content of other all-in-one yamls. (That's also why I reserved some ugly multiple-line spaces because I think we'd better not prune it by manual, for the complete automation.)

deploy/kubernetes/configmap.yaml
heypnus: hmm... in NCP the default value is indeed `k8scluster` for now. So I think if there's no upgrade issue for `cluster = k8scluster`, then there won't be upgrade issue for `#cluster = k8scluster`, do you think it makes sense?

deploy/kubernetes/configmap.yaml
dantingl: I just saw it's different with the options in openshift. I'm fine with current automatic way.

deploy/kubernetes/configmap.yaml
dantingl: make sense.

dantingl (approved):

deploy/kubernetes/configmap.yaml
timdengyun: May I ask that the value of entry commented out is the same with NCP default config value?

timdengyun (approved):

deploy/kubernetes/configmap.yaml
heypnus: > May I ask that the value of entry commented out is the same with NCP default config value?

yes

heypnus: I recovered to `adaptor = kubernetes` because the operator reported the error "failed to get key adaptor from section coe" in Kubernetes platform. This is because that we don't have `FillDefaults` for Kubernetes platform and the validation will check if the value exists. Some alternative solutions:
1. Add `FillDefaults` function for Kubernetes
2. Remove the valudation function for Kubernetes
3. Modify the render script in NCP to uncomment the value to make sure the generated yaml doesn't comment the value
I'll try 3 after the patch was merged.

PR #139, opened 2021-07-07
Force disable flag for waiting for security computing notification

salv-orlando:


manifest/kubernetes/rhel/ncp-rhel.yaml
```diff
@@ -639,7 +650,7 @@ data:
 # Yaml template for NCP Deployment
 # Proper kubernetes API and NSX API parameters, and NCP Docker image
 # must be specified.
-# This yaml file is part of NCP 3.1.1 release.
+# This yaml file is part of NCP 3.1.2 release.
```
dantingl, line 653: Do we need to set it to 3.1.1.1? I saw you also included some commits for 3.1.2

pkg/controller/configmap/config.go
```diff
@@ -80,7 +81,7 @@ func (adaptor *ConfigMapOc) FillDefaults(configmap *corev1.ConfigMap, spec *conf
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "adaptor", "openshift4", true))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "policy_nsxapi", "True", true))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "single_tier_topology", "True", true))
-	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "wait_for_security_policy_sync", "True", false))
+	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "wait_for_security_policy_sync", "False", true))
```
dantingl, line 84: Is it because NCP set wait_for_security_policy_sync True by default, so make it force disabled in Operator?

manifest/kubernetes/rhel/ncp-rhel.yaml
timdengyun: > Do we need to set it to 3.1.1.1? I saw you also included some commits for 3.1.2 manifest and version syncup

May I ask 3.1.1.1 is a branch or tag on 3.1.1 branch. Thanks. we assume operator and ncp have the same branch strategy from 3.1.2 onward.

salv-orlando: Frankly I do not have any idea about how this got updated to 3.1.2... yes, I'll set it to 3.1.1.1

pkg/controller/configmap/config.go
salv-orlando: I  wanted to make sure that no user can think "hey, let's try this, looks very interesting", and also - as you wrote - "fix" environments where the flag was set to True by forcing it to False.

manifest/kubernetes/rhel/ncp-rhel.yaml
salv-orlando: There is a problem in the way this PR was created! I need to force push it to make sure only my commit is included.

salv-orlando: Now it's better...

dantingl (approved):

PR #160, opened 2021-11-17
prepare NCP opeartor 3.2.0.1 release

timdengyun:
In preparation of NCP opeartor 3.2.0.1 release:
Update year to 2021 for LICENSE and Notice
Update VERSION to 3.2.0.1


LICENSE
```diff
@@ -1,6 +1,6 @@
 nsx-container-plugin-operator
 
-Copyright (c) 2020 VMware, Inc.  All rights reserved.				
+Copyright (c) 2021 VMware, Inc.  All rights reserved.
```
salv-orlando, line 3: I'm not sure if this needs to be updated for each release, usually we don't. But I don't think it'll do any harm updating it.

VERSION
```diff
@@ -1 +1 @@
-3.2.0-dev
+3.2.0.1
```
salv-orlando, line 1: thanks!

salv-orlando (approved):

LICENSE
timdengyun: Yes, I'm also not very sure on this update.  I did it as my previous experience.  Thanks for comments.

VERSION
timdengyun: Welcome. It's aligned with NCP release version.

Issue #54, opened 2020-09-09
Add node selector and tolerations as option in the configmap


yasensim:
Currently it is not possible to dictate where NCP will be deployed as the deployment spec is applied by the operator. Customers want to be able to run NCP on master, infra, or compute nodes. we need to provide this option either in the operator configmap or as an env var.

@dantingl  and @salv-orlando  can we plan it?

yasensim:
Also the daemon sets need to have configurable tolerations as many customers are using taints.

timdengyun:
Issue fixed.

timdengyun:
Hi gammohamed,
AS I know, there is no root password for ncp operator pod, you can access pod by using kubectl exec -it -n nsx-system-operator <operartor-pod-name> -- /bin/bash
but the file in the pod is not editable. 
If you want to change node selector and tolerations, you need to edit ncp-install CRD as in https://github.com/timdengyun/nsx-container-plugin-operator/commit/e6d7fe01e72c32703c662e9094102254c1b33674

PR #171, opened 2021-12-22
Add Ncp and Ds Tolerations from NcpInstall CRD

timdengyun:
This patch is to:
1. Allow user to configure tolerations and for NCP Deployment
and nsx-ncp-bootstrap/nsx-node-agent DaemonSet by editing
operator.nsx.vmware.com_v1_ncpinstall_cr.yaml spec field.
Operator dynamically detects user-defined spec changes to sync with
NCP Deployment and nsx-ncp-bootstrap/nsx-node-agent DaemonSet when user make changes.

You can add a taint to a node using kubectl taint. For example,
 kubectl taint nodes node1 key1=value1:NoSchedule
to places a taint on node node1.
The taint has key key1, value value1, and taint effect NoSchedule.
This means that no NCP pod will be able to schedule onto node1 unless it has a matching toleration.
So, you can specify a toleration below for NCP Deployment in the operator.nsx.vmware.com_v1_ncpinstall_cr.yaml

spec:
  nsx-ncp:
    tolerations:
      - key: "key1"
      operator: "Equal"
      value: "value1"
      effect: "NoSchedule"

The tolerations above "match" the taint created by the kubectl taint in the node1,
and thus NCP pod can be schedule onto node1, otherwise, NCP is in pending status
if node1 is selected to run NCP pod.

2. Move default tolerations of NCP Deployment and nsx-ncp-bootstrap,
nsx-node-agent DaemonSet from manifest yaml to operator.nsx.vmware.com_v1_ncpinstall_cr.yaml
CRD.

Resolves issue #54


pkg/controller/configmap/configmap_controller.go
```diff
@@ -630,7 +650,46 @@ func (r *ReconcileConfigMap) isNcpDeploymentChanged(ncpReplicas int32) (bool, er
 	}
 	prevImage := ncpDeployment.Spec.Template.Spec.Containers[0].Image
 	currImage := os.Getenv(operatortypes.NcpImageEnv)
-	if prevImage != currImage || ncpReplicas != *ncpDeployment.Spec.Replicas {
+
+	prevTolerations := ncpDeployment.Spec.Template.Spec.Tolerations
+	isTolerationsEqual := true
+	if ncpTolerations != nil {
+		isTolerationsEqual = reflect.DeepEqual(prevTolerations, *ncpTolerations)
+	}
```
dantingl, line 658: What if prevTolerations!=nil and ncpTolerations==nil?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -384,14 +389,28 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
...
 
+	// Check if nsx-node-agent and nsx-ncp-bootstrap DaemonSet chagned
+	nsxNodeAgentDsChanged := false
+	nsxNodeAgentDsChanged, err = r.isNsxNodeAgentDsChanged(&nsxNodeAgentDsTolerations)
+	if err != nil {
+		log.Error(err, "Failed to get nsxNodeAgentDsChanged")
+		return reconcile.Result{Requeue: true}, err
+	}
+
+	if !nsxNodeAgentDsChanged {
```
dantingl, line 400: Where do you use flag `nsxNodeAgentDsChanged`?

timdengyun: we use nsxNodeAgentDsChanged to print log info. Also, it's used in line 432 where we need to check if return or not when ncp deployment is not changed.

pkg/controller/configmap/configmap_controller.go
timdengyun: ncpTolerations should not be nil, it's not expected. as we discussed, since DeepEqual could check nil inside, we just pass ncpTolerations instead of *ncpTolerations

pkg/controller/configmap/configmap_controller.go
```diff
@@ -320,7 +324,8 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// Render configurations
-	objs, err := Render(instance, &ncpReplicas, opNsxSecret, opLbSecret)
+	objs, err := Render(instance, &ncpReplicas, &ncpTolerations, &nsxNodeAgentDsTolerations,
```
dantingl, line 327: out of curiosity, why put new parameters in the middle instead of in the end?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -409,7 +428,11 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 				}
 				r.status.SetNotDegraded(statusmanager.ClusterConfig)
 				r.status.SetNotDegraded(statusmanager.OperatorConfig)
-				return reconcile.Result{}, nil
+
+				if !nsxNodeAgentDsChanged {
```
dantingl, line 432: Do you think we should put this in the same level with `ncpDeploymentChanged` at line 418?

pkg/controller/configmap/configmap_controller.go
timdengyun: I just want to goup the parameters for the same resource together,  So, eventually, it calling looks like:
objs, err := Render(instance, &ncpReplicas, &ncpNodeSelector, &ncpTolerations, &nsxNodeAgentDsTolerations,
		opNsxSecret, opLbSecret)

&ncpReplicas, &ncpNodeSelector, &ncpTolerations are for NCP deployment.
nsxNodeAgentDsTolerations for node agent DaemonSet.
the last two is for secret. 

Also, the order of arguments passed in is kept the same one of handling logic inside function.

timdengyun:
Rebase after PR170  nodeSelector patch merged.

pkg/controller/configmap/configmap_controller.go
timdengyun: I think we should put nsxNodeAgentDsChanged check either upper lever or lower level at line 418 since each step is to to do further check to include all conditions. In the end, it check there is no any changes happen.

timdengyun: Yes, agreed. So I refactor code a little bit as at line 425:
 if !ncpDeploymentChanged && !nsxNodeAgentDsChanged {

dantingl (approved):

heypnus (approved):

PR #173, opened 2022-02-16
Set default coe adaptor for k8s env

timdengyun:
Since Container orchestrator adaptor option in configmap:
 'adaptor = kubernetes'
 is hidden, so it's needed to set default value 'kubernetes' 
of configmap in the code for k8s env.

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -56,6 +56,7 @@ func (adaptor *ConfigMapK8s) FillDefaults(configmap *corev1.ConfigMap, spec *con
 		log.Error(err, "failed to load ConfigMap")
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/configmap/config.go
```diff
@@ -56,6 +56,7 @@ func (adaptor *ConfigMapK8s) FillDefaults(configmap *corev1.ConfigMap, spec *con
 		log.Error(err, "failed to load ConfigMap")
 		return err
 	}
+	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "adaptor", "kubernetes", true))
```
timdengyun, line 59: fill default coe adaptor value since this option is hidden in configmap, the OC already did it in code line#82

heypnus (approved):

PR #174, opened 2022-02-17
Doc that flag "addNodeTag" should be disabled for node multiple NICs

heypnus:
If a node has multiple VirtualNetworkInterfaces attached, the operator
cannot detect which node LSP should be enabled. So the customer should
disable flag "addNodeTag" and tag the node LSP by manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,14 @@ Set container network CIDRs `clusterNetwork` in `$MY_CLUSTER/install-config.yaml
 ```
 $ openshift-install --dir=$MY_CLUSTER create manifests
 ```
+
+If one cluster node has multiple VirtualNetworkInterfaces, the operator cannot
+detect which interface should be enabled as the containers' parent interface,
+so the user should edit `deploy/openshift4/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml`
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -93,6 +93,13 @@ Set container network CIDRs `clusterNetwork` in `$MY_CLUSTER/install-config.yaml
 ```
 $ openshift-install --dir=$MY_CLUSTER create manifests
 ```
+
+If one cluster node has multiple VirtualNetworkInterfaces, the operator cannot
+detect which interface should be enabled as the containers' parent interface,
+so the user should edit `deploy/openshift4/operator.nsx.vmware.com_v1_ncpinstall_cr.yaml`
+to set `addNodeTag: false` and manually tag the target node port by
+`scope=ncp/node_name, tag=<node_name>` and `scope=ncp/node_name, tag=<cluster_name>`.
```
timdengyun, line 101: Just want to ask, where users go to tag node port, in NSXT side, is it? Thanks

timdengyun (approved):

README.md
heypnus: yes, in NSX-T. Thanks for reminder, added the info.

PR #187, opened 2022-04-27
sync ncp manifests with latest ncp

jwsui:
Update NCP manifest with latest NCP on branch master.

### DIFF
--- a/manifest/kubernetes/rhel/ncp-rhel.yaml
+++ b/manifest/kubernetes/rhel/ncp-rhel.yaml
@@ -36,6 +36,7 @@ spec:
         type: string
         description: The namespace of the k8s object if it is namespaced. None by default
         jsonPath: .spec.error-object-ns
+  preserveUnknownFields: false  
 
   scope: Cluster
   names:
@@ -62,6 +63,7 @@ spec:
         openAPIV3Schema:
           type: object
           x-kubernetes-preserve-unknown-fields: true
+  preserveUnknownFields: false
 
   scope: Namespaced
   names:
@@ -86,6 +88,7 @@ spec:
         openAPIV3Schema:
           type: object
           x-kubernetes-preserve-unknown-fields: true
+  preserveUnknownFields: false
 
   scope: Namespaced
   names:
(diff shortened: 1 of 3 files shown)


manifest/kubernetes/rhel/ncp-rhel.yaml
```diff
@@ -950,27 +960,6 @@ spec:
           - name: host-var-run
             mountPath: /var/run/openvswitch
             subPath: openvswitch
-{{if .UseNsxOvsKmod}}
```
dantingl, line 953: Please talk with @2ez4szliu if this is fine to be removed, not sure if he already added this part in NCP yaml, so Operator could sync it automatically.

jwsui: ok, checked with latest NCP master, this part has not been added.

timdengyun: I think we still need to keep all sections of .UseNsxOvsKmod since they are still needed in operator manifest. 
once shizhao fix in NCP side, we will have same code here, just generated code not handy-input code.

manifest/openshift4/coreos/ncp-openshift4.yaml
```diff
@@ -452,6 +457,57 @@ data:
 
 
 
+---
```
timdengyun, line 460: This part should be removed because it's for ocp3. I have not change it in NCP upstream not to generate it so we have to manually delete it each time when syncing manifests.

timdengyun:
We need to make changes according in NCP upstream manifest templates as well to generate manifest yaml automatically

manifest/kubernetes/rhel/ncp-rhel.yaml
jwsui: Got it, this part has been added.

manifest/openshift4/coreos/ncp-openshift4.yaml
jwsui: Thanks and fixed.

timdengyun (approved):

PR #192, opened 2022-06-06
Fixing the cursor issue while achieving node

heypnus:
The default page size of NSX-T API is 1000. In the scale env,
when the number of nodes is greater than 1000, we need to specify
parameter cursor in multiple queries to get all node information.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -117,18 +117,27 @@ func getNodeExternalIdByProviderId(nsxClients *NsxClients, nodeName string, prov
 	if len(providerId) != 46 {
 		return "", errors.Errorf("invalid provider ID %s of node %s", providerId, nodeName)
 	}
+	localVarOptionals := make(map[string]interface{})
 	providerId = string([]byte(providerId)[10:])
 	nsxClient := nsxClients.ManagerClient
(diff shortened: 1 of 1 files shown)


dantingl (approved):

timdengyun:
func listAttachmentsByNodeExternalId(nsxClients *NsxClients, vmExternalId string) ([]string, error) {
	var attachment_ids []string
	localVarOptionals := make(map[string]interface{})
	localVarOptionals["ownerVmId"] = vmExternalId
	nsxClient := nsxClients.ManagerClient
vifs, _, err := nsxClient.FabricApi.ListVifs(nsxClient.Context, localVarOptionals)

Here, we also use the same way to list vifs of vms, do we also need make similar changes?

heypnus:
> func listAttachmentsByNodeExternalId(nsxClients *NsxClients, vmExternalId string) ([]string, error) { var attachment_ids []string localVarOptionals := make(map[string]interface{}) localVarOptionals["ownerVmId"] = vmExternalId nsxClient := nsxClients.ManagerClient vifs, _, err := nsxClient.FabricApi.ListVifs(nsxClient.Context, localVarOptionals)
> 
> Here, we also use the same way to list vifs of vms, do we also need make similar changes?

Thanks for your comment. I think we don't need it in the other lists because we had filters(ownerVmId for vifs, attachmentId for ports) in the other invokes, I believe that the number of returned values won't be 1000+

PR #193, opened 2022-06-08
Add configmap option diff change log

timdengyun:
This patch is to enhance log output when user makes configmap change:
1.Log which section is added into or removed from configmap
2.Log configmap option added/uncommented or removed/commented
3.Log configmap option value change

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -445,6 +445,7 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	if prevConfig == nil {
 		return podNeedChange{true, true, true}, nil
 	}
+
 	currData := currConfig.Data
 	prevData := prevConfig.Data
 	// Compare the whole data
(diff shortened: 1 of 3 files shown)


pkg/controller/configmap/config.go
```diff
@@ -469,11 +470,35 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
...
+				curKeyValue := currCfg.Section(name).Key(key).Value()
+				if prevCfg.Section(name).HasKey(key) {
+					preKeyValue := prevCfg.Section(name).Key(key).Value()
+					if !reflect.DeepEqual(preKeyValue, curKeyValue) {
+						log.Info(fmt.Sprintf("Section [%s] config option: %s = %s is changed to %s = %s",
+							name, key, preKeyValue, key, curKeyValue))
+					}
+				} else {
+					log.Info(fmt.Sprintf("Section [%s] add/uncomment a new config option: %s = %s",
+						name, key, curKeyValue))
```
dantingl, line 491: Do you think we should hide the value as maybe it's related to cert/password?

timdengyun: From debug perspective, we want to show some option value's in the log for better understanding what changes have been done. We might change it to debug level log. but printing value itself in somehow is worthy. Or we can
do special handing for cert/password to hide value. That's maybe better.

dantingl (approved):

PR #195, opened 2022-06-13
Add operator log in support bundle

timdengyun:
Nsxcli collect pod logs by using "tier: nsx-networking" label
to select matching pod. This patch is to add "tier: nsx-networking"
label in ncp-operator pod deployment yaml definition to make nsxcli
can collect operator log.

### DIFF
--- a/deploy/kubernetes/operator.yaml
+++ b/deploy/kubernetes/operator.yaml
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: nsx-ncp-operator
   namespace: nsx-system-operator
+  labels:
+    tier: nsx-networking
 spec:
   replicas: 1
   selector:
@@ -12,6 +14,7 @@ spec:
     metadata:
       labels:
         name: nsx-ncp-operator
+        tier: nsx-networking
     spec:
       hostNetwork: true
       serviceAccountName: nsx-ncp-operator
--- a/deploy/openshift4/operator.yaml
+++ b/deploy/openshift4/operator.yaml
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: nsx-ncp-operator
   namespace: nsx-system-operator
+  labels:
+    tier: nsx-networking
 spec:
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/operator.yaml
```diff
@@ -3,15 +3,23 @@ kind: Deployment
 metadata:
   name: nsx-ncp-operator
   namespace: nsx-system-operator
```
dantingl, line 5: Add to OC manifest as well?

timdengyun: Added. Thanks for reminder.

deploy/kubernetes/operator.yaml
```diff
@@ -12,6 +16,9 @@ spec:
     metadata:
       labels:
         name: nsx-ncp-operator
+        tier: nsx-networking
```
timdengyun, line 19: we must Add "tier: nsx-networking" labels to spec.metadata so that these labels could be applied to pod deployment.

deploy/kubernetes/operator.yaml
```diff
@@ -3,6 +3,10 @@ kind: Deployment
 metadata:
   name: nsx-ncp-operator
   namespace: nsx-system-operator
+  labels:
+    tier: nsx-networking
+    component: nsx-ncp-operator
+    version: v1
 spec:
   replicas: 1
   selector:
```
timdengyun, line 12: we shouldn't change
 matchLabels:
   name: nsx-ncp-operator

since matchLabels are the immutable filed, it can not be directly updated for running pods, so, we have to delete old pods to make it effective. but it will make rolling update failed.

timdengyun:
> If only label `tier` matters, `component` and `version` labels can be removed?



> If only label `tier` matters, `component` and `version` labels can be removed?

Yes, `component` and `version` labels can be removed. At this moment, we don't find their value.

dantingl (approved):

PR #202, opened 2022-09-02
Ensure the error message can be removed after node removal

heypnus:
This patch will fix the error in the following case:

There's an error while tagging the node port, the operator reports
the error to cluster operator, then the node is removed from the
cluster. If the error is the last one in the cluster operator, it
cannot be cleaned as expected.


dantingl (approved):

salv-orlando (approved):
LGTM!

PR #204, opened 2022-10-10
Use set to avoid the duplicate vifs when searching by owner_vm_id

heypnus:
In some corner case, according to the response of vif search API, there can be 2 vifs for the same VM, which including the same lport_attachment_id and MAC. To avoid store the same attachent id in the variable attachment_ids, we change the var type from list to set.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -35,6 +35,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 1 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -157,7 +158,7 @@ func listAttachmentsByNodeExternalId(nsxClients *NsxClients, vmExternalId string
 	}
 	for _, vif := range vifs.Results {
 		if vif.LportAttachmentId != "" {
-			attachment_ids = append(attachment_ids, vif.LportAttachmentId)
+			attachment_ids.Insert(vif.LportAttachmentId)
```
timdengyun, line 161: Current solution certainly can fix duplicated LportAttachmentId case. 
however,from search output, we see there two vis associated with the same node. I see one has invalid IpAddressInfo and other has empty IpAddressInfo. I think the latter one with empty IpAddressInfo is from power-offed vm. So, do we need to check which vis is valid using IpAddressInfo  info.

timdengyun:
> In some corner case, according to the response of vif search API, there can be 2 vifs for the same VM, which including the same lport_attachment_id and MAC. To avoid store the same attachent id in the variable attachment_ids, we change the var type from list to set.

which including the same lport_attachment_id and MAC -> "which includes the same lport_attachment_id and MAC", is it better?

pkg/controller/node/node_controller.go
heypnus: It's safe because we don't read the address from the vifs, we read it from K8s side. Our target is to get the node LSP id. The 2 vifs have the same attachment id, and there is only one port refering to the attachment id. In the other words, the vif attachment and the logic port is unique, that's enough. The logic flow is: get IP and provider id from K8s side -> get node external id by provider id -> get attachment ids by node external id -> get logic ports by attachment ids -> filter the logic ports by IP, then the result should be a single node LSP.

timdengyun: Thanks for detailed depiction. It should be ok to skip IpAddressInfo check.

timdengyun (approved):

PR #208, opened 2022-12-23
Reset configmap log_file option

timdengyun:
The log_file config option in DEFAULT section is used for both nsx-ncp-config and nsx-node-agent-config.
So, the log_file will be used for bootstrap, ncp and node-agent pods once it is set. All pods will write log in the same log_file. However, ncp and node-agent write log_file with ncp user, bootstrap pod write log_file with root user.

The log_file is set in operator configmap, boostrap, ncp and node-agent pod will be restarted. If bootstrap pod create log_file before than ncp/node-agent pods is trying to access. Bootstrap pod creates and write log_file with root users. In such case, the ncp/node-agent will hit Permission denied issue when writing log_file and the ncp/node-agent pod will be crashed.

This patch is to make a workaround solution to fix this issue:
1. Reset log_file to the default None value so that ncp, bootstrap and node-agent will use their own default logfile, respectively.
2. Print a log message when detecting log_file option is set by user.

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -105,8 +105,8 @@ func (adaptor *ConfigMapOc) FillDefaults(configmap *corev1.ConfigMap, spec *conf
 
 func (adaptor *ConfigMapK8s) validateConfigMap(configmap *corev1.ConfigMap) []error {
 	errs := []error{}
-	data := configmap.Data
-	cfg, err := ini.Load([]byte(data[operatortypes.ConfigMapDataKey]))
+	data := &configmap.Data
+	cfg, err := ini.Load([]byte((*data)[operatortypes.ConfigMapDataKey]))
 	if err != nil {
 		errs = append(errs, errors.Wrapf(err, "failed to load ConfigMap"))
 		return errs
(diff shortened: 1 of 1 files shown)


pkg/controller/configmap/config.go
```diff
@@ -158,6 +168,16 @@ func (adaptor *ConfigMapOc) validateConfigMap(configmap *corev1.ConfigMap) []err
 	if err != nil {
 		appendErrorIfNotNil(&errs, errors.Wrapf(err, "mtu is invalid"))
 	}
+	// Check DEFAULT section log_file option
```
seanpang-vmware, line 171: this part is the same as 132-141, maybe add a method instead of same code block.

timdengyun: We trying to merge the two func (adaptor *ConfigMapOc) validateConfigMap(configmap *corev1.ConfigMap) and func (adaptor *ConfigMapK8s) validateConfigMap(configmap *corev1.ConfigMap) since the logic of these two functions are basically the same. but it's a little bit rush to do it in the operator 4.1.0. We can consider this in the next release.

heypnus (approved):

pkg/controller/configmap/config.go
```diff
@@ -129,15 +129,25 @@ func (adaptor *ConfigMapK8s) validateConfigMap(configmap *corev1.ConfigMap) []er
...
+	if cfg.Section("DEFAULT").Key("log_file").Value() != "" {
+		log.Info(fmt.Sprintf("DEFAULT section option: log_file is not allowed to change, current value:%s is reverted back to default value:None",
+			cfg.Section("DEFAULT").Key("log_file").Value()))
+		// Revert DEFAULT log_file to default value
+		fillDefault(cfg, "DEFAULT", "log_file", "", true)
+		// Write config back to ConfigMap data
+		(*data)[operatortypes.ConfigMapDataKey], err = iniWriteToString(cfg)
```
salv-orlando, line 139: Do we need to reset the value call iniWriteToString if in any case we are failing validation and therefore setting operator in degraded state?

timdengyun: if there is failing or error when calling iniWriteToString, we write error in NCP install CRD status, so, it means NCP install CRD is a degraded state. I feel this approach should be ok.

PR #214, opened 2023-03-14
Add enhancement for ncp-operator config

jwsui:
Make ncp-operator responsible for inferring adaptor by os-release of host node, and remove coe.adaptor from ConfigMap. Add check for immutable fields to prevent these fields from being changed.

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -29,6 +29,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+var immutableFields = []struct {
+	Section string
+	Key     string
+}{
+	{
+		Section: "coe",
+		Key:     "cluster",
+	},
+	{
+		Section: "nsx_v3",
+		Key:     "l4_lb_auto_scaling",
+	},
+}
+
 type Adaptor interface {
 	FillDefaults(configmap *corev1.ConfigMap, spec *configv1.NetworkSpec) error
 	validateConfigMap(configmap *corev1.ConfigMap) []error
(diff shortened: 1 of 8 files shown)


pkg/controller/configmap/config.go
```diff
@@ -571,6 +578,29 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	return needChange, nil
 }
 
+func immutableFieldChanged(cur, prev *ini.File) bool {
+	immutableFields := []struct {
+		Section string
+		Key     string
+	}{
```
dantingl, line 585: Do we need a const dict to save the options?

pkg/controller/sharedinfo/shared_info.go
```diff
@@ -40,33 +40,35 @@ type SharedInfo struct {
 	NsxNcpDeploymentSpec  *unstructured.Unstructured
 }
 
+func getAdaptorName() (string, error) {
+	osReleasePath, envSet := os.LookupEnv("OS_RELEASE_PATH")
+	if !envSet {
+		osReleasePath = "/host/etc/os-release"
```
dantingl, line 46: OsReleaseFile in https://github.com/vmware/nsx-container-plugin-operator/blob/52d9f351cc7812a7c3878833453640638a478df9/pkg/types/names.go#LL41C2-L41C15

jwsui: Thanks, done.

pkg/controller/configmap/config.go
jwsui: Golang does not support constant arrays, maps or slices, and I think we can use global variable instead.

dantingl (dismissed):

dantingl (approved):

PR #215, opened 2023-03-16
Enhance configmap option validation

timdengyun:
1. Hide options below in configmap yaml:
nsx_api_cert_file, nsx_api_private_key_file, ca_file in nsx_v3 section
lb_default_cert_path, lb_priv_key_path in nsx_v3 section
policy_nsxapi in nsx_v3 section since only policy api is supported
ovs_bridge in nsx_node_agent section since it's an advanced option
top_firewall_section_marker and bottom_firewall_section_marker in nsx_v3 section,
these two options are MP specific options.

2. Automate nsx authentication configuration options:
nsx_api_cert_file, nsx_api_private_key_file and ca_file
when user creates nsx-secret using
https://github.com/vmware/nsx-container-plugin-operator/blob/main/deploy/kubernetes/nsx-secret.yaml
Operator will auto fill these three parameters in configmap based on tls.crt,
tls.key and tls.ca fields in nsx-secret:
nsx_api_cert_file = /etc/nsx-ujo/nsx-cert/tls.crt
nsx_api_private_key_file = /etc/nsx-ujo/nsx-cert/tls.key
ca_file = /etc/nsx-ujo/nsx-cert/tls.ca

3. Automate lb certification configuration options:
lb_default_cert_path and lb_priv_key_path
when user creates lb-secret using
https://github.com/vmware/nsx-container-plugin-operator/blob/main/deploy/kubernetes/lb-secret.yaml
Operator will auto fill these two parameters in configmap based on tls.crt,
tls.key fields in lb-secret:
lb_default_cert_path = /etc/nsx-ujo/lb-cert/tls.crt
lb_priv_key_path = /etc/nsx-ujo/lb-cert/tls.key

4. Reject MP specific options once user sets.

5. Reject WCP specific options once user sets.

6. Reject TAS specific options once user sets.

7. Set policy_nsxapi as True for K8s platform since operator only
supports Policy API mode.


deploy/kubernetes/configmap.yaml
```diff
@@ -174,7 +174,7 @@ data:
     # no baseline rule will be created and the cluster will assume the default
     # behavior as specified by the backend. The option is only supported on
     # Policy API.
-    # Choices: <None> allow_cluster allow_namespace allow_namespace_strict
+    # Choices: allow_cluster allow_namespace allow_namespace_strict None NONE none
```
dantingl, line 177: `None NONE none`: I guess we can just list one

deploy/kubernetes/configmap.yaml
```diff
@@ -368,7 +355,7 @@ data:
     # 'DENY' will enable logging only for DENY rules. Remove this config if no
     # logging is desired. When IPv6 is enabled this setting will not apply to
     # rules for allowing ND traffic.
-    # Choices: ALL DENY <None>
+    # Choices: ALL DENY None NONE none
```
dantingl, line 358: Same here

pkg/controller/configmap/config.go
```diff
@@ -137,22 +137,12 @@ func (adaptor *ConfigMapK8s) validateConfigMap(configmap *corev1.ConfigMap) []er
...
-		// Write config back to ConfigMap data
-		(*data)[operatortypes.ConfigMapDataKey], err = iniWriteToString(cfg)
-		appendErrorIfNotNil(&errs, err)
-	}
+
+	errs = append(errs, validateNodeAgentOptions(configmap, cfg)...)
+	errs = append(errs, validateCommonOptions(configmap, cfg)...)
+	errs = append(errs, validateMPOptions(configmap, cfg)...)
+	errs = append(errs, validateWCPOptions(configmap, cfg)...)
+	errs = append(errs, validateTASOptions(configmap, cfg)...)
```
dantingl, line 145: Let's do not distinguish other platforms, we can put them all in one blocklist.

pkg/controller/configmap/config.go
```diff
@@ -166,6 +156,7 @@ func (adaptor *ConfigMapOc) validateConfigMap(configmap *corev1.ConfigMap) []err
 		errs = append(errs, errors.Wrapf(err, "failed to load ConfigMap"))
 		return errs
 	}
+	appendErrorIfNotNil(&errs, validateConfig(cfg, "coe", "adaptor"))
```
dantingl, line 159: Remove this line?

pkg/controller/configmap/config.go
```diff
@@ -316,6 +297,102 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+	return errs
+}
+
+// Validate common options for both ncp and node agent
+func validateCommonOptions(configmap *corev1.ConfigMap, cfg *ini.File) []error {
+	errs := []error{}
+	data := &configmap.Data
+	var err error
+	// Check DEFAULT section log_file option
+	if cfg.Section("DEFAULT").Key("log_file").Value() != "" {
```
dantingl, line 319: Could you help explain why log_file cannot be changed

pkg/controller/configmap/config.go
```diff
@@ -316,6 +297,102 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+	}
+	return errs
+}
+
+// Validate MP specific options
+func validateMPOptions(configmap *corev1.ConfigMap, cfg *ini.File) []error {
+	errs := []error{}
+	data := &configmap.Data
+	var err error
+	if cfg.Section("nsx_v3").Key("policy_nsxapi").Value() == "True" {
```
dantingl, line 336: Put section and key into a map, so we can also add new options in the future

pkg/controller/configmap/config.go
```diff
@@ -316,6 +297,102 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+		if cfg.Section("nsx_v3").Key("top_firewall_section_marker").Value() != "" {
+			log.Info(fmt.Sprintf("nsx_v3 section option: top_firewall_section_marker is not supported in PolicyAPI, reverted to default value:None"))
+			appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "top_firewall_section_marker", "", true))
+		}
+		if cfg.Section("nsx_v3").Key("bottom_firewall_section_marker").Value() != "" {
+			log.Info(fmt.Sprintf("nsx_v3 section option: bottom_firewall_section_marker is not supported in PolicyAPI, reverted to default value:None"))
+			appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "bottom_firewall_section_marker", "", true))
+		}
+
+		// Write config back to ConfigMap data
```
dantingl, line 346: Why we need to revert the config? Will Operator do reconcile all the time when hitting this kind of error?
I thought we just need to mark Operator as degraded and do not apply the new change.

pkg/controller/configmap/config.go
```diff
@@ -733,3 +812,86 @@ func IsMTUChanged(currConfigMap *corev1.ConfigMap, prevConfigMap *corev1.ConfigM
...
+			appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "lb_default_cert_path", "/etc/nsx-ujo/lb-cert/tls.crt", true))
+		}
+	}
+
+	if lb_key, found := lbSecret.Data["tls.key"]; found {
+		if len(lb_key) == 0 {
+			log.Info(fmt.Sprintf("tls.key is set as None in %s", operatortypes.LbSecret))
+			appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "lb_priv_key_path", "", true))
+		} else {
+			appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "lb_priv_key_path", "/etc/nsx-ujo/lb-cert/tls.key", true))
```
dantingl, line 885: Put `"/etc/nsx-ujo/lb-cert/tls.key"` in defaults?

deploy/kubernetes/configmap.yaml
timdengyun: These three values are from auto generated configmap yaml based recent ncp changes.So, it might be to keep as is

deploy/kubernetes/configmap.yaml
timdengyun: These three values are from auto generated configmap yaml based recent ncp changes.So, it might be to keep as is

pkg/controller/configmap/config.go
timdengyun: You mean merge validateWCPOptions and validateWCPOptions into one function?

pkg/controller/configmap/config.go
timdengyun: I want to add this line to check if coe.adaptor is set or not by FillDefaults function. Jianwei's patch will automatically check env to put adaptaor info in shared status. and then to decide create k8s or ocp4 Adapter instance.

pkg/controller/configmap/config.go
timdengyun: revert back to default value, and run make operator running as normal not degraded status. In fact, these options are hidden in configmap, as usual, user should not set these options,  Do you think mark operator as degraded is better?

pkg/controller/configmap/config.go
timdengyun: I'm thinking the ncp will hit lb certification error if the tls.key is empty in the secret, however, we still set lb_priv_key_path as /etc/nsx-ujo/lb-cert/tls.key.

pkg/controller/configmap/config.go
timdengyun: good suggestion, thanks

pkg/controller/configmap/config.go
timdengyun: This is a workaround solution. Currently, ncp configmap and node agent configmap will use the same option for log_file. If user don't set log_file, the ncp and node agent will default log file name. The log file are different for ncp and node agent log. Once user set log_file option, ncp and node 
the log_file will be used for bootstrap, ncp and node agent pods. All pods will write log in the same log_file. However, ncp and node agent write log_file with ncp user, bootstrap pod
write log_file with root user. when bootstrap pod create log_file before than ncp/node-agent pods is trying to create it.  The log_file file permission will be set
-rw-r--r--.  1 root root     9186 Dec 13 03:11 test_logfile
It means other uses other than root can not write this file any more. So, In such case, the ncp/node-agent will hit Permission denied issue when writing log_file.

pkg/controller/configmap/config.go
dantingl: Yes, call it like validateConfigOptions. I feel it's better do not expose other platforms in operator code.

pkg/controller/configmap/config.go
```diff
@@ -316,6 +297,102 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+func validateWCPOptions(configmap *corev1.ConfigMap, cfg *ini.File) []error {
+	errs := []error{}
+	data := &configmap.Data
+	is_changed := false
+	var err error
+
+	for section, keys := range operatortypes.WcpOptions {
+		for _, key := range keys {
+			if optionInConfigMap(configmap, section, key) {
+				log.Info(fmt.Sprintf("%s section option: %s is WCP specific option and not supported, deleted this config option", section, key))
```
heypnus, line 363: For the unsupported options, I think we don't have to explain what we will handle it, delete option or revert to default, etc. I personally prefer to just log "it will be ignored".

pkg/controller/configmap/config.go
```diff
@@ -733,3 +812,86 @@ func IsMTUChanged(currConfigMap *corev1.ConfigMap, prevConfigMap *corev1.ConfigM
...
+func FillNsxAuthCfg(configmap *corev1.ConfigMap, nsxSecret *corev1.Secret) error {
+	errs := []error{}
+	data := &configmap.Data
+	cfg, err := ini.Load([]byte((*data)[operatortypes.ConfigMapDataKey]))
+	if err != nil {
+		log.Error(err, "failed to load ConfigMap")
+		return err
+	}
+
+	if nsx_cert, found := nsxSecret.Data["tls.crt"]; found {
```
heypnus, line 825: So if the "tls.crt" was not found in nsxSecret, i.e. found is false, we don't need to do anything, right?

pkg/controller/configmap/config.go
timdengyun: Agreed. Thanks, I will log it as ignored

pkg/controller/configmap/config.go
timdengyun: yes, I think so. In this case, user should use adim/password for auth, of course, these should not in production env, might be in test env.

pkg/controller/configmap/config.go
timdengyun: Agreed, we reject MP, WCP and TAS specific options and don't revert these options when user sets. And for DEFAULT.log_file, we don't revert it back any more, as suggested, we log an error and mark ncp install as degraded

pkg/controller/configmap/config.go
dantingl: I don't understand why we still need to check it as it is set by operator itself?
I thought Jianwei removed checking env

pkg/controller/configmap/config.go
```diff
@@ -316,6 +293,78 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+
+// Validate non supported options
+// Including WCP and TAS specific options, MP supported options
+func validateUnSupportedOptions(configmap *corev1.ConfigMap) []error {
+	errs := []error{}
+	data := &configmap.Data
+	is_changed := false
+	cfg, err := ini.Load([]byte((*data)[operatortypes.ConfigMapDataKey]))
+
+	if cfg.Section("nsx_v3").Key("policy_nsxapi").Value() == "True" {
```
dantingl, line 332: I guess operator only supports policy_nsxapi==True?

timdengyun: Agreed, we can remove this check here, since policy_nsxapi should be default as True.

pkg/controller/configmap/config.go
timdengyun: I get your point, we indeed don't need validate coe.adatpore anymore, since it's set by operator instead of user.

pkg/controller/configmap/config.go
```diff
@@ -71,6 +71,7 @@ func (adaptor *ConfigMapK8s) FillDefaults(configmap *corev1.ConfigMap, spec *con
 		return err
 	}
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "coe", "adaptor", "kubernetes", true))
```
jwsui, line 73: Just want to confirm that whether we should remove this line?  @timdengyun @dantingl 
I didn't remove this in my patch as I thought `coe.adaptor` should be filled and then rendered to ncp configmap.

pkg/controller/configmap/config.go
heypnus: Then can we simplify L793-L800 to:
```
	if nsx_cert, found := nsxSecret.Data["tls.crt"]; found && len(nsx_cernt) > 0 {
		appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_v3", "nsx_api_cert_file", "/etc/nsx-ujo/nsx-cert/tls.crt", true))
	}
```
?
Because I think `None` is the default value so that we don't need to set it explicitly?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -307,23 +310,34 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// Get nsx-secret and lb-secret
-	var opNsxSecret *corev1.Secret
-	if optionInConfigMap(instance, "nsx_v3", "nsx_api_cert_file") {
-		opNsxSecret = &corev1.Secret{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: watchedNamespace, Name: operatortypes.NsxSecret}, opNsxSecret)
+	opNsxSecret := &corev1.Secret{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: watchedNamespace, Name: operatortypes.NsxSecret}, opNsxSecret)
```
heypnus, line 314: Is it safe to remove `if optionInConfigMap(instance, "nsx_v3", "nsx_api_cert_file") {...` before L314?

pkg/types/defaults.go
```diff
@@ -8,7 +8,30 @@ const (
...
 
-var NcpSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-var AgentSections = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
-var OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
-var BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+var (
+	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
+	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
+	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
```
heypnus, line 15: It seems that we don't need section `vc` anymore?

timdengyun: yes, that is true, we can remove vc section

pkg/controller/configmap/configmap_controller.go
timdengyun: yes, I think so, there should be no nsx_api_cert_file option any more since these options are removed from configmap.

pkg/controller/configmap/config.go
```diff
@@ -316,6 +291,76 @@ func validateConfig(cfg *ini.File, sec string, key string) error {
...
+				cfg.Section(section).DeleteKey(key)
+				is_changed = true
+			}
+		}
+	}
+
+	for section, keys := range operatortypes.WCPOptions {
+		for _, key := range keys {
+			if optionInConfigMap(configmap, section, key) {
+				log.Info(fmt.Sprintf("%s section option: %s is WCP specific option and not supported, ignored this config option", section, key))
```
dantingl, line 343: Again I feel it's not a good idea to expose any other platform in log or code, could we remove WCP here?
Maybe we can just merge MPOptions and WCPOptions to UnsupportedOptions?

timdengyun: agreed. I remove these words from log but keep in code.

pkg/controller/configmap/config.go
timdengyun: I tested again. I agreed with you advice.  we don't need to set it explicitly when secret field is empty.  if secret is empty, this option should not be set by user as well. otherwise, there will be some issue. So, we just need to check the case that secret is created, and we set it for users. This will be safe for both fresh installation and upgrade

heypnus (approved):
LGTM

PR #216, opened 2023-03-21
Correct unsupported options

timdengyun:
Fix unsupported options in correct sections:
node_type is in coe section
network_info_resync_period is in k8s section


pkg/types/defaults.go
```diff
@@ -31,7 +31,10 @@ var WCPOptions = map[string][]string{
 	"k8s": {
 		"enable_vnet_crd", "enable_lb_monitor_crd", "enable_nsxnetworkconfig_crd", "enable_routeset_crd",
 		"enable_ip_pool_crd", "enable_vm_crd", "lb_statistic_monitor_interval", "enable_lb_vs_statistics_monitor",
-		"ip_usage_alarm_threshold", "node_type",
+		"network_info_resync_period", "ip_usage_alarm_threshold",
+	},
+	"coe": {
+		"node_type",
```
dantingl, line 37: We do not need set node_type for running Operator either on K8s or OC, right?

dantingl (approved):

pkg/types/defaults.go
timdengyun: yes, we don't node_type for k8s or OCP.

PR #217, opened 2023-04-14
Avoid unnecessary nsx-node-agent restarts

jwsui:
It is not necessary to restart nsx-node-agent when some keys are changed in section `k8s` and `coe`. So update the code to only restart nsx-node-agent when specified keys in section `k8s` and `coe` are changed.

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -562,8 +562,7 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	diffSecs := []string{}
 	currSecs := currCfg.SectionStrings()
 	for _, name := range currSecs {
-		_, err = prevCfg.GetSection(name)
-		if err != nil {
+		if !hasSection(prevCfg, name) {
 			if len(currCfg.Section(name).KeyStrings()) != 0 {
 				log.Info(fmt.Sprintf("Section [%s] is added into configmap", name))
 				diffSecs = append(diffSecs, name)
@@ -613,8 +612,7 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	}
 	prevSecs := prevCfg.SectionStrings()
 	for _, name := range prevSecs {
-		_, err = currCfg.GetSection(name)
-		if err != nil {
+		if !hasSection(currCfg, name) {
 			if len(prevCfg.Section(name).KeyStrings()) != 0 {
 				log.Info(fmt.Sprintf("Section [%s] is removed from configmap", name))
 				diffSecs = append(diffSecs, name)
@@ -630,7 +628,10 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 		if !needChange.agent && inSlice(sec, operatortypes.AgentSections) {
 			needChange.agent = true
 		}
-		if !needChange.bootstrap && inSlice(sec, operatortypes.BootstrapOptions) {
+		if !needChange.agent && checkOptionsChange(operatortypes.AgentOptionKeys, sec, prevCfg, currCfg) {
+			needChange.agent = true
+		}
+		if !needChange.bootstrap && checkOptionsChange(operatortypes.BootstrapOptionKeys, sec, prevCfg, currCfg) {
 			needChange.bootstrap = true
 		}
 		if needChange.ncp && needChange.agent && needChange.bootstrap {
@@ -643,6 +644,26 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	}
 
 	return needChange, nil
+}
+
+func checkOptionsChange(options map[string][]string, sec string, prevCfg, currCfg *ini.File) bool {
+	if _, ok := options[sec]; !ok {
+		return false
+	}
+	if hasSection(prevCfg, sec) != hasSection(currCfg, sec) {
+		return true
+	}
+	// We can assert that section `sec` exists in both prevCfg and currCfg here.
+	for _, key := range options[sec] {
+		if prevCfg.Section(sec).HasKey(key) != currCfg.Section(sec).HasKey(key) {
+			return true
+		}
+		if prevCfg.Section(sec).HasKey(key) && currCfg.Section(sec).HasKey(key) &&
+			prevCfg.Section(sec).Key(key).Value() != currCfg.Section(sec).Key(key).Value() {
+			return true
+		}
+	}
+	return false
 }
 
 func immutableFieldChanged(cur, prev *ini.File) bool {
@@ -729,6 +750,11 @@ func GenerateOperatorConfigMap(opConfigmap *corev1.ConfigMap, ncpConfigMap *core
 		return err
 	}
 	return nil
+}
+
+func hasSection(cfg *ini.File, section string) bool {
+	_, err := cfg.GetSection(section)
+	return err == nil
 }
 
 func iniWriteToString(cfg *ini.File) (string, error) {
--- a/pkg/types/defaults.go
+++ b/pkg/types/defaults.go
@@ -9,10 +9,26 @@ const (
 )
 
 var (
-	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
-	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
-	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+	NcpSections         = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
+	AgentSections       = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
+	OperatorSections    = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
+	BootstrapOptionKeys = map[string][]string{
+		"DEFAULT": {
+			"log_dir", "log_file", "log_rotation_file_max_mb", "log_rotation_backup_count",
+		},
+		"nsx_node_agent": {
+			"enable_ipv6", "use_nsx_ovs_kernel_module", "ovs_db_sock", "ovs_bridge", "ovs_uplink_port", "mtu",
+		},
+		"coe": {
+			"node_type",
+		},
+	}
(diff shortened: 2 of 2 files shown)


pkg/controller/configmap/config.go
```diff
@@ -623,6 +621,24 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 		if !needChange.agent && inSlice(sec, operatortypes.AgentSections) {
 			needChange.agent = true
 		}
+		if !needChange.agent && inSlice(sec, []string{"k8s", "coe"}) {
```
timdengyun, line 624: operatortypes.AgentSections are []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}, which means any changes in one of section AgentSections, needChange.agent will be set true at #L622.  If this, the logic from #L624 won't be processed. 
Here, we only want to check what option changes will make node agent restart. So, we want to restart node agent ds only when changes are in AgentKeys, is it?

pkg/controller/configmap/config.go
```diff
@@ -623,6 +621,24 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
...
+			if hasSection(prevCfg, sec) != hasSection(currCfg, sec) {
```
timdengyun, line 625: In most cases, k8s can code sections are compulsory, so previous config and current config will have both.

pkg/controller/configmap/config.go
timdengyun: we may not need code line#L621-622

pkg/controller/configmap/config.go
```diff
@@ -623,6 +621,24 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
...
+				needChange.agent = true
+			} else {
+				// We can assert that section `sec` exists in both prevCfg and currCfg here.
+				for _, key := range operatortypes.AgentKeys {
+					if prevCfg.Section(sec).HasKey(key) != currCfg.Section(sec).HasKey(key) {
```
timdengyun, line 630: if changed key is not from AgentKeys, do we need to restart node agent as well? if not, we need to check added key or removed key from AgentKeys or not to decide if restarting node agent or not?

pkg/types/defaults.go
```diff
@@ -13,6 +13,7 @@ var (
 	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
 	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
 	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+	AgentKeys        = []string{"apiserver_host_ip", "apiserver_host_port", "client_token_file", "ca_file", "enable_multus", "enable_hostport_snat", "connect_retry_timeout"}
```
timdengyun, line 16: AgentKeys -> AgentRestartKeys maybe more clear.

pkg/controller/configmap/config.go
jwsui: > operatortypes.AgentSections are []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}, which means any changes in one of section AgentSections, needChange.agent will be set true at #L622. If this, the logic from #L624 won't be processed. Here, we only want to check what option changes will make node agent restart. So, we want to restart node agent ds only when changes are in AgentKeys, is it?

Actually we need line#L621-622, but I forget to change defaults.go. Please check the latest change in defaults.go, in this file, section k8s and coe are removed. This means that we only need to check certain keys of this two sections. While we still need to keep other sections, so we still need 621-622.

pkg/controller/configmap/config.go
jwsui: > In most cases, k8s can code sections are compulsory, so previous config and current config will have both.

But we still need a check here as the section could be `coe`?

pkg/controller/configmap/config.go
jwsui: > if changed key is not from AgentKeys, do we need to restart node agent as well? if not, we need to check added key or removed key from AgentKeys or not to decide if restarting node agent or not?

Not really get you point, but I think this judgement is to restart node-agent when there is added or removed key?

pkg/types/defaults.go
jwsui: > AgentKeys -> AgentRestartKeys maybe more clear.

Thanks, done.

pkg/types/defaults.go
```diff
@@ -10,9 +10,10 @@ const (
 
 var (
 	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentSections    = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
 	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
 	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+	AgentRestartKeys = []string{"apiserver_host_ip", "apiserver_host_port", "client_token_file", "ca_file", "enable_multus", "enable_hostport_snat", "connect_retry_timeout"}
```
timdengyun, line 16: it's better to change this as a map as we did for WCPOptions. The good point of map is to better know the map relationship between section and option.

pkg/controller/configmap/config.go
timdengyun: Thanks for explanation.

pkg/controller/configmap/config.go
timdengyun: I see your point. So you only want to restart node agent when key is from AgentRestartKeys, and the other key in k8s and coe are be ignored no matter what changes are made.

jwsui: Yes, that's right.

pkg/types/defaults.go
jwsui: > it's better to change this as a map as we did for WCPOptions. The good point of map is to better know the map relationship between section and option.

Thanks, let me check.

pkg/controller/configmap/config.go
timdengyun: Yes, agreed, if adding or deleting a section, like coe or k8s, restarting node agent might be necessary. More precisely,  we can check if added or deleted section has key from AgentRestartKey or not.

pkg/types/defaults.go
dantingl, line 16: No need to watch multus option

dantingl, line 16: `AgentRestartKeys` -> AgentOptionKeys

pkg/types/defaults.go
```diff
@@ -10,9 +10,10 @@ const (
 
 var (
 	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentSections    = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
 	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
 	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
```
dantingl, line 15: Do we need to recheck `BootstrapOptions`, does it have impact on node networking?

pkg/types/defaults.go
```diff
@@ -10,9 +10,10 @@ const (
 
 var (
 	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentSections    = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
```
dantingl, line 13: We should not remove k8s right, you need to compare apiserver_host_ip in AgentRestartKeys

dantingl: Another way I'm thinking is to only check options instead of checking sections

jwsui: In current implementation, we need to remove `k8s` and `coe` section, and I added key like `apiserver_host_ip` in `AgentRestartKeys`. This means that only certain key like `apiserver_host_ip` change will trigger operator to restart node-agent.

pkg/types/defaults.go
jwsui: Agree that map will make it more clear, so update options with structure map, thanks.

jwsui: ok, it's removed.

jwsui: done

pkg/types/defaults.go
jwsui: Good point, we'd better have a check. Code updated.

ksamoray:
@jwsui rebase this change so github actions can run

pkg/types/defaults.go
```diff
@@ -9,10 +9,26 @@ const (
...
+			"log_dir", "log_file", "log_rotation_file_max_mb", "log_rotation_backup_count",
+		},
+		"nsx_node_agent": {
+			"enable_ipv6", "use_nsx_ovs_kernel_module", "ovs_db_sock", "ovs_bridge", "ovs_uplink_port", "mtu",
+		},
+		"coe": {
+			"node_type",
+		},
+	}
+	AgentOptionKeys = map[string][]string{
```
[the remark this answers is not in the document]
jwsui: > Is it possible to identify DEFAULT section options to restart node agent like we discovered for BootstrapOptionKeys, the motivation here is to minimize possibility of restarting of node agent

I think node-agent requires all DEFAULT section options in current node-agent implementation.

jwsui:
> @jwsui rebase this change so github actions can run

Thanks for you advice, all checks have passed.

pkg/types/defaults.go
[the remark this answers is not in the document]
timdengyun: I see.

timdengyun (dismissed):

pkg/types/defaults.go
```diff
@@ -9,10 +9,26 @@ const (
...
+		"k8s": {
+			"apiserver_host_ip", "apiserver_host_port", "client_token_file", "ca_file", "enable_multus", "enable_hostport_snat", "connect_retry_timeout",
```
timdengyun, line 28: one more suggest, we will remove multus code soon, I suggest we remove enable_multus here. Later, we will also remove it from configmap.

jwsui: Thanks for reminder, it's removed.

timdengyun (approved):

PR #219, opened 2023-04-20
Check if sharedInfo returned err before use

ksamoray:
Check the returned status from sharedInfo initialization before it is used, as otherwise it fails ungracefully.


timdengyun (approved):
LGTM, Thanks for correction.

PR #222, opened 2023-04-23
Add node info in identifying node agents to be restarted

timdengyun:


pkg/controller/pod/pod_controller.go
```diff
@@ -334,8 +337,8 @@ func identifyPodsInCLBDueToInvalidResolvConf(c client.Client) (
 				nodeAgentLogs, "Failed to establish a new connection: "+
 					"[Errno -2] Name or service not known") {
 				log.Info(fmt.Sprintf(
-					"Pod %v is in CLB because of invalid resolv.conf. "+
-						"It shall be restarted", pod.Name))
+					"Pod %v in node %s is in CLB because of invalid resolv.conf. "+
```
dantingl, line 340: Why pod use %v, node use %s?

PR #225, opened 2023-05-29
Avoid unnecessary nsx-node-agent restarts

jwsui:
It is not necessary to restart nsx-node-agent when some keys are changed in section `k8s` and `coe`. So update the code to only restart nsx-node-agent when specified keys in section `k8s` and `coe` are changed.

### DIFF
--- a/pkg/types/defaults.go
+++ b/pkg/types/defaults.go
@@ -10,9 +10,26 @@ const (
 
 var (
 	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
 	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
-	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+	// AgentSections are responsible for rendering nsx-node-agent configmap, while not all keys updates in section `coe` and
+	// `k8s` require restart of nsx-node-agent. So AgentRestartSections are responsible for checking whether nsx-node-agent
+	// should be restarted when related keys are updated.
+	AgentSections              = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentRestartSections       = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
+	BootstrapRestartOptionKeys = map[string][]string{
+		"DEFAULT": {
+			"log_dir", "log_file", "log_rotation_file_max_mb", "log_rotation_backup_count",
+		},
+		"nsx_node_agent": {
+			"enable_ipv6", "use_nsx_ovs_kernel_module", "ovs_uplink_port", "mtu",
+		},
+	}
+	AgentRestartOptionKeys = map[string][]string{
+		"k8s": {
+			"apiserver_host_ip", "apiserver_host_port", "client_token_file", "ca_file", "enable_hostport_snat",
+		},
+		"coe": {"connect_retry_timeout"},
+	}
 )
 
 var TASSection = string("cf")
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -562,8 +562,7 @@ func NeedApplyChange(currConfig *corev1.ConfigMap, prevConfig *corev1.ConfigMap)
 	diffSecs := []string{}
 	currSecs := currCfg.SectionStrings()
 	for _, name := range currSecs {
-		_, err = prevCfg.GetSection(name)
-		if err != nil {
+		if !hasSection(prevCfg, name) {
 			if len(currCfg.Section(name).KeyStrings()) != 0 {
 				log.Info(fmt.Sprintf("Section [%s] is added into configmap", name))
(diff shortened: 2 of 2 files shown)


pkg/types/defaults.go
```diff
@@ -9,10 +9,27 @@ const (
 )
 
 var (
-	NcpSections      = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
-	AgentSections    = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
-	OperatorSections = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
-	BootstrapOptions = []string{"DEFAULT", "nsx_node_agent"}
+	NcpSections                = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc"}
+	AgentSections              = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
```
timdengyun, line 13: Yes, I just noted that we need to keep AgentSections since the sections in AgentSections are needed to render operator config map. in Render function:
renderData.Data[operatortypes.NodeAgentConfigMapRenderKey], err = generateConfigMap(cfg, operatortypes.AgentSections)
So, it's good to add some comments to explain why we put AgentSections here, considering we also have AgentRestartSections and AgentRestartOptionKeys.

pkg/types/defaults.go
```diff
@@ -9,10 +9,27 @@ const (
...
+	OperatorSections           = []string{"DEFAULT", "ha", "k8s", "coe", "nsx_v3", "vc", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentRestartSections       = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
+	BootstrapRestartOptionKeys = map[string][]string{
+		"DEFAULT": {
+			"log_dir", "log_file", "log_rotation_file_max_mb", "log_rotation_backup_count",
+		},
+		"nsx_node_agent": {
+			"enable_ipv6", "use_nsx_ovs_kernel_module", "ovs_db_sock", "ovs_bridge", "ovs_uplink_port", "mtu",
+		},
+		"coe": {
```
timdengyun, line 23: why we add "coe":{node_type} for BootstrapRestartOptionKeys since we see previous BootstrapOptions only have DEFAULT and nsx_node_agent section?

jwsui: Adding this because it's used in `init_k8s_node` in bootstrap, so I thinks we need add this field unless the field is not allowed to be updated.

> why we add "coe":{node_type} for BootstrapRestartOptionKeys since we see previous BootstrapOptions only have DEFAULT and nsx_node_agent section?

pkg/types/defaults.go
jwsui: Done

pkg/types/defaults.go
timdengyun: I noted that "coe":{node_type} is an advanced option, which means is not visible to users. 

cfg.StrOpt('node_type',
               choices=[constants.HOSTVM_NODE_TYPE, constants.WCP_WORKER],
               default=constants.HOSTVM_NODE_TYPE,
               advanced=True,
               help='The type of container host node'),
               
we only run once   init_k8s_node in  bootstrap, during start, after then, init_container exit.    

So，I think users don't need to update this option after running.

pkg/types/defaults.go
```diff
@@ -10,9 +10,29 @@ const (
...
+			"node_type",
+		},
+	}
+	AgentRestartOptionKeys = map[string][]string{
+		"k8s": {
+			"apiserver_host_ip", "apiserver_host_port", "client_token_file", "ca_file", "enable_hostport_snat", "connect_retry_timeout",
```
timdengyun, line 32: I don't see "connect_retry_timeout" in k8s sections https://github.com/vmware/nsx-container-plugin-operator/blob/main/deploy/openshift4/configmap.yaml
only see it in coe section, do we need put it in k8s?

jwsui: Thanks for clarify, so `coe.node_type` and `k8s.connect_timeout` could be removed.

pkg/types/defaults.go
```diff
@@ -10,9 +10,26 @@ const (
...
+	// `k8s` require restart of nsx-node-agent. So AgentRestartSections are responsible for checking whether nsx-node-agent
+	// should be restarted when related keys are updated.
+	AgentSections              = []string{"DEFAULT", "k8s", "coe", "nsx_node_agent", "nsx_kube_proxy"}
+	AgentRestartSections       = []string{"DEFAULT", "nsx_node_agent", "nsx_kube_proxy"}
+	BootstrapRestartOptionKeys = map[string][]string{
+		"DEFAULT": {
+			"log_dir", "log_file", "log_rotation_file_max_mb", "log_rotation_backup_count",
+		},
+		"nsx_node_agent": {
+			"enable_ipv6", "use_nsx_ovs_kernel_module", "ovs_db_sock", "ovs_bridge", "ovs_uplink_port", "mtu",
```
timdengyun, line 24: we see ovs_db_sock and ovs_bridge are advanced options as well, so we can skip these two options as well since they are not exposed to users. I think we only need to handle the options visible to users.

jwsui: Indeed, I didn't realize that some options are advanced. Had removed advanced options. Thanks.

timdengyun (approved):

PR #227, opened 2023-06-09
Sync NCP 4.1.1 manifests/configmap yamls

timdengyun:
Sync manifests/configmap yamls from NCP 4.1.1
nsx-keeper-411 branch to operator main branch


jwsui (approved):
LGTM.

VERSION
```diff
@@ -1 +1 @@
-4.1.0-dev
+4.1.1
```
heypnus, line 1: Why it's not 4.1.1-dev? I think `4.1.1` should only be used in the branch `release_4.1.1`,  right?
btw, not related to this patch, I found the version in https://github.com/vmware/nsx-container-plugin-operator/blob/release_4.0.1/VERSION in not correct.. it should be `4.0.1`

timdengyun: make sense. I just want to simplify version change between main branch and release branch. So, I opt to change it as 4.1.1. However, given our tradition version control procedure, it could be 4.1.1-dev. 

Thanks for bring a hidden issue, I think it should be an overlook change during release 4.0.1. I gonna fix it in a new patch.

timdengyun: Fix it. Thanks

PR #234, opened 2023-10-08
Sync manifest and configmap change of NCP 4.1.2

timdengyun:
Removing PodSecurityPolicy(PSP) since K8s 1.25 no longer doesn't support PSP

Removing volMount /usr/lib64/libcrypto.so.1.1 in bootstrap pod for OCP4 to support OCP 4.13

### DIFF
--- a/deploy/kubernetes/configmap.yaml
+++ b/deploy/kubernetes/configmap.yaml
@@ -83,6 +83,7 @@ data:
     #enable_sha = True
 
 
+
     [ha]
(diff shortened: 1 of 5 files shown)


2ez4szliu:
LGTM
I think the change for supporting openshift 4.13 is already merge, not sure if we need to sync the fork branch with main branch before merging.

2ez4szliu (approved):

PR #241, opened 2023-11-27
Upgrade actions/checkout, actions/setup-go

ksamoray:
Versions is use are about to be deprecated, upgrade to latest versions of the above.


ksamoray:
`cache: false` addresses another warn by GitHub actions, as we have no go.sum file for the project.

timdengyun (approved):

PR #250, opened 2024-03-26
Set enable_ovs_mcast_snooping as false by default

dantingl:
keepalived is configured with unicast by default from OC 4.11

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -110,8 +110,9 @@ func (adaptor *ConfigMapOc) FillDefaults(configmap *corev1.ConfigMap, spec *conf
 	// For Openshift add a 3-seconds agent delay by default
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "waiting_before_cni_response", "3", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "mtu", strconv.Itoa(operatortypes.DefaultMTU), false))
-	// For Openshift force enable_ovs_mcast_snooping as True by default for IPI and UPI installation
-	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "enable_ovs_mcast_snooping", "true", false))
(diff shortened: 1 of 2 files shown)


pkg/controller/configmap/config.go
```diff
@@ -110,8 +110,9 @@ func (adaptor *ConfigMapOc) FillDefaults(configmap *corev1.ConfigMap, spec *conf
 	// For Openshift add a 3-seconds agent delay by default
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "waiting_before_cni_response", "3", false))
 	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "mtu", strconv.Itoa(operatortypes.DefaultMTU), false))
-	// For Openshift force enable_ovs_mcast_snooping as True by default for IPI and UPI installation
-	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "enable_ovs_mcast_snooping", "true", false))
+	// For Openshift force enable_ovs_mcast_snooping as False by default for IPI and UPI installation
+        // keepalived is configured with unicast by default from OC 4.11
+	appendErrorIfNotNil(&errs, fillDefault(cfg, "nsx_node_agent", "enable_ovs_mcast_snooping", "false", false))
```
timdengyun, line 115: Do we change force from false to true?So, it will always reset enable_ovs_mcast_snooping to false no matter what value user sets in the configmap.If we still keep this option, it might not be necessary to do it.

timdengyun (approved):
LGTM, just a minor comment

pkg/controller/configmap/config.go
dantingl: No, I think keepalived can still be configured with multicast. We just need to follow the same default behavior and still allow user to use multicast way.
It is necessary to do this, so user does not need to set the flag manually as it is kind of transparent for the default behavior.

timdengyun: As I know, multicast no longer support multicast instead of unicast. Of course, we can keep just keep the default setting value in accordance with default keepalived default behaviour.

PR #260, opened 2024-10-03
Switch to Policy API for Segment and Segment Ports

ggverma:
As NSX 9.0 intends to remove MP APIs, we need to switch to Policy API. This patch makes the changes necessary for node controller Logical Switch related APIs

The patch removes some stale code which is no longer used

Testing Done
------------
- Verified that the Segment Ports for nodes are updated by the operator
- Verified that the operator is running fine

Jira: NCP-601

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,8 +14,8 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
 	github.com/vmware/go-vmware-nsxt v0.0.0-20201207175959-23201aae9cc3
-	github.com/vmware/vsphere-automation-sdk-go/runtime v0.2.0
-	github.com/vmware/vsphere-automation-sdk-go/services/nsxt v0.3.0
+	github.com/vmware/vsphere-automation-sdk-go/runtime v0.7.0
+	github.com/vmware/vsphere-automation-sdk-go/services/nsxt v0.12.0
 	gopkg.in/ini.v1 v1.51.0
 	k8s.io/api v0.18.2
 	k8s.io/apimachinery v0.18.3
@@ -45,11 +45,12 @@ require (
 	github.com/go-logr/logr v0.1.0 // indirect
 	github.com/go-logr/zapr v0.1.1 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
+	github.com/golang-jwt/jwt/v4 v4.3.0 // indirect
 	github.com/golang/groupcache v0.0.0-20191027212112-611e8accdfc9 // indirect
 	github.com/golang/protobuf v1.5.0 // indirect
 	github.com/google/go-cmp v0.5.5 // indirect
 	github.com/google/gofuzz v1.1.0 // indirect
-	github.com/google/uuid v1.1.1 // indirect
+	github.com/google/uuid v1.2.0 // indirect
 	github.com/googleapis/gnostic v0.3.1 // indirect
 	github.com/gophercloud/gophercloud v0.6.0 // indirect
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
@@ -66,7 +67,7 @@ require (
 	github.com/prometheus/client_model v0.2.0 // indirect
 	github.com/prometheus/common v0.26.0 // indirect
 	github.com/prometheus/procfs v0.6.0 // indirect
-	github.com/vmware/vsphere-automation-sdk-go/lib v0.2.0 // indirect
+	github.com/vmware/vsphere-automation-sdk-go/lib v0.7.0 // indirect
 	go.uber.org/atomic v1.6.0 // indirect
 	go.uber.org/multierr v1.5.0 // indirect
 	go.uber.org/zap v1.14.1 // indirect
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -17,17 +17,16 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	nsxt "github.com/vmware/go-vmware-nsxt"
-	"github.com/vmware/go-vmware-nsxt/common"
-	nsxtmgr "github.com/vmware/go-vmware-nsxt/manager"
 
 	"github.com/vmware/nsx-container-plugin-operator/pkg/controller/sharedinfo"
 	"github.com/vmware/nsx-container-plugin-operator/pkg/controller/statusmanager"
 	operatortypes "github.com/vmware/nsx-container-plugin-operator/pkg/types"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/core"
-	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
 	vspherelog "github.com/vmware/vsphere-automation-sdk-go/runtime/log"
 	policyclient "github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/security"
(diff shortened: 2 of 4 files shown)


pkg/types/utils.go
```diff
@@ -45,3 +48,25 @@ func identifyAndGetInstance(resName string) (runtime.Object, error) {
 	}
 	return nil, errors.Errorf("failed to identify instance for: %s", resName)
 }
+
+func CastToBindingType[T any](dataValue *data.StructValue, destBindingType bindings.BindingType) (T, error) {
+	converter := bindings.NewTypeConverter()
+	obj, errs := converter.ConvertToGolang(dataValue, destBindingType)
+	if len(errs) > 0 {
+		t, _ := obj.(T)
```
salv-orlando, line 56: Is there any need for assigning `t` here? I think you can probably just return `nil, errs[0]`

pkg/controller/node/node_controller.go
```diff
@@ -595,15 +540,26 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
...
-		nsxClient := nsxClients.ManagerClient
-		_, _, err = nsxClient.LogicalSwitchingApi.UpdateLogicalPort(nsxClient.Context, lsp.Id, *lsp)
+		connector := nsxClients.PolicyConnector
+		segPortStateClient := segSDK.NewPortsClient(connector)
+		segId, err := operatortypes.ExtractSegmentIdFromPath(*lsp.ParentPath)
+		if err != nil {
+			cachedNodeSet[nodeName] = &statusmanager.NodeStatus{
+				Addresses: nodeAddresses,
+				Success:   false,
+				Reason:    fmt.Sprintf("Error while achieving port with specific address for node %s: %v", nodeName, err),
```
salv-orlando, line 557: I know this error message has been like this for year, do you mind improving it?
It goes into the node status and not just in the logs.

Maybe just use "retrieving" instead of "achieving" (and do this for all the place where we ser this error in status)

pkg/controller/node/node_controller.go
```diff
@@ -197,23 +200,28 @@ func inSlice(str string, s []string) bool {
...
 			continue
 		}
 		var collectedAddresses []string
-		for _, realizedBinding := range logicalPort.RealizedBindings {
-			address := realizedBinding.Binding.IpAddress
+		for _, realizedBinding := range portState.RealizedBindings {
+			address := *realizedBinding.Binding.IpAddress
 			if inSlice(address, nodeAddresses) && !inSlice(address, collectedAddresses) {
-				log.Info(fmt.Sprintf("Node address %s matches port %s", address, port.Id))
+				log.Info(fmt.Sprintf("Node address %s matches port %s", address, *port.Id))
```
salv-orlando, line 224: nit: change this line to `node address %s matches segment port %s`.
In this way, when we will check logs for troubleshooting, we will know that the operator is executing the code for policy API.
Also, I think the linter suggests to start log lines with a lowercase letter, so it might be another good thing to do.

pkg/controller/node/node_controller.go
```diff
@@ -197,23 +200,28 @@ func inSlice(str string, s []string) bool {
...
-		logicalPort, _, err := nsxClient.LogicalSwitchingApi.GetLogicalPortState(nsxClient.Context, port.Id)
+		segId, err := operatortypes.ExtractSegmentIdFromPath(*port.ParentPath)
 		if err != nil {
 			return filteredPorts, err
 		}
-		if len(logicalPort.RealizedBindings) == 0 {
+		portState, err := segPortStateClient.Get(segId, *port.Id, nil, nil)
+		if err != nil {
+			return filteredPorts, err
+		}
```
salv-orlando, line 216: For this check and the one at line 210 I'm thinking that perhaps we want to do thing a little differently:
Instead of returning at the first error, keep iterating over ports, and then return slices for FilteredPorts and errors.

The thing is that we want to find "the" port with the right IP, so even if we encounter issues while fetching or parsing state for other ports, we might still be able to find the port we are looking for.

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
 	github.com/vmware/go-vmware-nsxt v0.0.0-20201207175959-23201aae9cc3
-	github.com/vmware/vsphere-automation-sdk-go/runtime v0.2.0
-	github.com/vmware/vsphere-automation-sdk-go/services/nsxt v0.3.0
+	github.com/vmware/vsphere-automation-sdk-go/runtime v0.7.0
```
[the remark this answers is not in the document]
ggverma: runtime has v0.7.0 as latest. These don't have to be the same, I guess

https://github.com/vmware/vsphere-automation-sdk-go/tags

pkg/types/utils.go
ggverma: can't return nil as per the compiler

ggverma: but I found a better way to do this

pkg/controller/node/node_controller.go
ggverma: done

pkg/controller/node/node_controller.go
ggverma: done

go.mod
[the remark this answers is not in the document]
salv-orlando: then it does not matter, ignore this!

pkg/controller/node/node_controller.go
ggverma: what we can do alternatively is finish the loop collecting errors. If there are any ports which couldn't be read from NSX, we return 1 error providing the info of all erroneous ports

salv-orlando: sounds good!

salv-orlando (approved):
LGTM

pkg/controller/node/node_controller.go
timdengyun, line 1: I think we should remove using SDK: 
nsxt "github.com/vmware/go-vmware-nsxt"
so, we should use nsx manger client at code line#L106
https://github.com/vmware/nsx-container-plugin-operator/pull/260/files#diff-92f10e93215b8da12723d913e0985568114aab4b8afb67541c2a3a9ec5cafba6R106

timdengyun, line 1: we need to use some inventory API but they are supported as well to list vif and vms
Here, we can use this client from sdk:
https://github.com/zhengxiexie/vsphere-automation-sdk-go/blob/services/nsxt/v0.12.0/services/nsxt-mp/nsx/fabric/VifsClient.go 
to avoid using nsxt "github.com/vmware/go-vmware-nsxt"

timdengyun, line 1: ditto

ggverma: hey Yun, the change you suggest is doable but I am not sure if it matches the criteria for this patch as the intention here is to switch those MP api calls to Policy api calls which are going to be removed. The suggested change switches the sdk but still uses mp api and these api calls are not planned for removal.

ggverma: same as above

timdengyun: Yes, I agreed. It's good to hear from advice from @salv-orlando advice on the way we go. If there is no strong option to change sdk, we can skip this removal and check in the patch.

salv-orlando: I advise to remove the deprecated go-vmware-nsxt client and all its usages in a different PR.
Let's use this PR exclusively for switching logical ports and switches to policy API.

timdengyun (approved):
LGTM

PR #262, opened 2024-11-05
Allow NcpReplicas to be set to 0

salv-orlando:
There are some scenarios were NCP must be temporarily shut down. This can be done by setting the number of replicas to 0, but the operator currently prevents that.


VERSION
```diff
@@ -1 +0,0 @@
-4.2.1
```
salv-orlando, line 1: This file should not have been removed by this PR. I will fix this.

timdengyun (approved):
LGTM, Thanks

PR #263, opened 2024-12-13
Enhance nsx-secret and lb-secert handling

timdengyun:
Skip handing nsx-secret and lb-secert when they are not found in the system.
This enhancment will allow users not set nsx-secret and lb-secert, but NCP Operator will not be stuck in keeping fetch them. So, the operator can continue to reconcile process instead of fetching loop.


salv-orlando (approved):
LGTM!

salv-orlando (approved):
LGTM

PR #264, opened 2024-12-25
Node Controller: Skip filtering by IP when node has only 1 port

salv-orlando:
Once the NodeController identifies node ports, the controller filters them by IP address to identify which ports need to be tagged. This is to ensure only the port which will be used for container networking has hyperbus mode enabled.

This commit skips IP filtering when only one port is configured for the node, as this step is not needed. This will also prevent issues in scenarios where node IPs are not correctly reported in NSX ports' realized address bindings.


dantingl (approved):

timdengyun (approved):
LGTM

PR #265, opened 2024-12-27
Update NCP ClusterRole to access ingresscontrollers

yuntanghsu:
ncp-openshift4.yaml is outdated and it needs to be update in order to access ingresscontrollers from api group operator.openshift.io for NCP

### DIFF
--- a/manifest/openshift4/coreos/ncp-openshift4.yaml
+++ b/manifest/openshift4/coreos/ncp-openshift4.yaml
@@ -224,6 +224,16 @@ rules:
 
 
 
+
+ - apiGroups:
+   - operator.openshift.io
+   resources:
+   - ingresscontrollers
+   verbs:
+   - list
+
+
+
 ---
 
 # Create ClusterRole for NCP to edit resources


salv-orlando (approved):
LGTM!

manifest/openshift4/coreos/ncp-openshift4.yaml
```diff
@@ -224,6 +224,16 @@ rules:
 
 
 
+
+ - apiGroups:
```
timdengyun, line 228: Do we have the patch in NCP manifest template as well, usually, we want to generate all platform manifests and configmap from template?

yuntanghsu: Yes, we generated the whole file when building sb/ob. I'm not sure if want to update the whole file for this change? I thought we will do batch update whenever a version is released?

timdengyun: Yes, We will do a bath update of the whole manifest file once release branch is cut.

timdengyun (approved):

PR #279, opened 2025-07-29
Sync OC manifest for running d-bus commands on host

ggverma:
Sync OC manifest for removing redundant volume mounts


timdengyun (approved):
LGTM, Thanks @ggverma

PR #309, opened 2026-06-29
build: move to two-stage Docker build, upgrade to Go 1.26.4

salv-orlando:
- Activate the golang-build stage in Dockerfile to compile the operator binary inside Docker (CGO_ENABLED=0)
- Upgrade base image from UBI8 to UBI9 minimal
- Upgrade Go toolchain from 1.26.2 to 1.26.4 (CVE fixes)
- Remove pre-build step from Makefile: binary is now built entirely within the Docker multi-stage build

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,5 +1,15 @@
-FROM golang:1.26.2 as golang-build
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM golang:1.26.4 AS golang-build
+
+WORKDIR /workspace
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+ARG LDFLAGS=""
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "${LDFLAGS}" -o build/bin/nsx-ncp-operator ./cmd/manager
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 
 LABEL name="nsx-container-plugin-operator"
 LABEL maintainer="NSX Containers Team <container-dev.pdl@broadcom.com>"
@@ -15,11 +25,11 @@ ENV OPERATOR=/usr/local/bin/nsx-ncp-operator \
     USER_UID=1001 \
     USER_NAME=nsx-ncp-operator
 
-COPY build/bin /usr/local/bin
+COPY build/bin/entrypoint /usr/local/bin/entrypoint
+COPY build/bin/user_setup /usr/local/bin/user_setup
+COPY --from=golang-build /workspace/build/bin/nsx-ncp-operator /usr/local/bin/nsx-ncp-operator
 COPY manifest /manifest
-RUN  rpm -e --nodeps curl-minimal || true && \
-     rpm -e --nodeps libcurl-minimal || true  && \
-     /usr/local/bin/user_setup
+RUN  /usr/local/bin/user_setup
 
 ENTRYPOINT ["/usr/local/bin/entrypoint"]
 
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/vmware/nsx-container-plugin-operator/pkg/apis"
 	operatorv1 "github.com/vmware/nsx-container-plugin-operator/pkg/apis/operator/v1"
+	"github.com/vmware/nsx-container-plugin-operator/version"
 )
 
 const (
@@ -93,7 +94,6 @@ func TestOperatorE2E(t *testing.T) {
 	if err := rbacv1.AddToScheme(scheme); err != nil {
 		t.Fatalf("Failed to add rbac APIs to scheme: %v", err)
 	}
-
 	k8sClient, err := client.New(config, client.Options{Scheme: scheme})
 	if err != nil {
 		t.Fatalf("Failed to create controller-runtime client: %v", err)
@@ -121,6 +121,32 @@ func TestOperatorE2E(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Operator deployment failed to become ready: %v", err)
 		}
+	})
+
+	// --- Version Reporting ---
+	t.Run("Operator Reports Correct Version", func(t *testing.T) {
+		pods, err := clientset.CoreV1().Pods(operatorNamespace).List(ctx, metav1.ListOptions{
+			LabelSelector: "name=nsx-ncp-operator",
+		})
+		if err != nil || len(pods.Items) == 0 {
+			t.Fatalf("Failed to list operator pods: %v", err)
+		}
+
+		expectedLine := fmt.Sprintf("Operator Version: %s", version.Version)
+		var logBytes []byte
+		err = wait.PollImmediate(2*time.Second, 30*time.Second, func() (bool, error) {
+			logBytes, err = clientset.CoreV1().Pods(operatorNamespace).
+				GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).
+				DoRaw(ctx)
+			if err != nil {
+				return false, err
+			}
+			return strings.Contains(string(logBytes), expectedLine), nil
+		})
+		if err != nil {
+			t.Fatalf("Operator pod logs do not contain %q — version stamp is missing or wrong.\nLogs:\n%s", expectedLine, string(logBytes))
+		}
+		t.Logf("Operator pod logs confirm version %s", version.Version)
 	})
 
 	// Setup: Patch existing nodes with vsphere ProviderIDs so the node controller can reconcile them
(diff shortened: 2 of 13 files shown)


build/Dockerfile
```diff
@@ -1,5 +1,10 @@
-FROM golang:1.26.2 as golang-build
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM golang:1.26.4 AS golang-build
+
```
ggverma, line 2: `go build` is missing `-ldflags`, so `version.Version`, `version.GitSHA`, and `version.GitTreeState` (injected via `versioning.mk`) will be empty strings in every binary built through Docker. The old Makefile `build` target passed `-ldflags '$(LDFLAGS)'`; this stage does not.

Suggested fix:
```dockerfile
ARG LDFLAGS=""
RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "${LDFLAGS}" -o build/bin/nsx-ncp-operator ./cmd/manager
```
and pass `--build-arg LDFLAGS="$(LDFLAGS)"` from the Makefile.

ggverma (commented):
## Review findings

Five issues, two of them silent regressions in the correctness of the produced binary.

build/Dockerfile
```diff
@@ -1,5 +1,10 @@
...
+WORKDIR /workspace
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o build/bin/nsx-ncp-operator ./cmd/manager
```
ggverma: `go build` is missing `-ldflags`, so `version.Version`, `version.GitSHA`, and `version.GitTreeState` (injected by `versioning.mk`) will be empty strings in every binary produced by this Docker stage. The old Makefile `build` target passed `-ldflags '$(LDFLAGS)'`; this `RUN` does not.

Suggested fix:
```dockerfile
ARG LDFLAGS=""
RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "${LDFLAGS}" -o build/bin/nsx-ncp-operator ./cmd/manager
```
and in the Makefile: `docker build --build-arg LDFLAGS="$(LDFLAGS)" -f build/Dockerfile ...`

build/Dockerfile
```diff
@@ -1,5 +1,10 @@
-FROM golang:1.26.2 as golang-build
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM golang:1.26.4 AS golang-build
+
+WORKDIR /workspace
+COPY . .
```
ggverma, line 7: `COPY . .` with no `.dockerignore` sends the entire repository as Docker build context, including `ncp_op.tar` and any previously built binary in `build/bin/`. This also embeds `.git/` in the builder stage layer.

The pattern used in `cmd/nsx-mock/Dockerfile` (same repo) is better: copy only `go.mod`/`go.sum` first, run `go mod download`, then copy source. This preserves the module-download layer cache across source edits. Without it, every change to any file (including YAML manifests) re-downloads all Go modules.

At minimum, add a `.dockerignore` excluding `build/bin/nsx-ncp-operator`, `ncp_op.tar`, `.git/`, and test artifacts.

build/Dockerfile
```diff
@@ -15,7 +20,9 @@ ENV OPERATOR=/usr/local/bin/nsx-ncp-operator \
     USER_UID=1001 \
     USER_NAME=nsx-ncp-operator
 
-COPY build/bin /usr/local/bin
+COPY build/bin/entrypoint /usr/local/bin/entrypoint
+COPY build/bin/user_setup /usr/local/bin/user_setup
+COPY --from=golang-build /workspace/build/bin/nsx-ncp-operator /usr/local/bin/nsx-ncp-operator
 COPY manifest /manifest
 RUN  rpm -e --nodeps curl-minimal || true && \
```
ggverma: `ubi9/ubi-minimal` does not ship the `rpm` binary — only `microdnf`. These two `rpm -e` lines will exit 127 and be silently swallowed by `|| true`, so `curl-minimal` and `libcurl-minimal` remain installed. The security hardening intent of these lines is defeated.

Replace with:
```dockerfile
RUN microdnf remove -y curl-minimal libcurl-minimal && microdnf clean all
```

ggverma:
**Makefile `build-test-images` target (line 64): local `go build` is now dead work**

The `build-test-images` target still runs:
```makefile
CGO_ENABLED=0 GOOS=linux $(GO) build -o $(BINDIR)/$(OPERATOR_NAME) $(GOFLAGS) -ldflags '$(LDFLAGS)' ./cmd/manager
```
…before calling `docker build -f build/Dockerfile`. Since the Dockerfile now compiles the binary in the `golang-build` stage, the locally produced `build/bin/nsx-ncp-operator` is never copied into the image — the stage rebuilds from source. The local build step is wasted time and produces a binary with correct ldflags that silently gets replaced by an unversioned one (see the separate comment on the `RUN go build` line).

This line should be removed from `build-test-images`, and the ldflags issue addressed in the Dockerfile instead.

build/Dockerfile
salv-orlando: These commands were not only not working because rpm is not available, they were also not working because we can't remove libcurl-minimal.

I scanned CVEs for libcurl-minimal and concluded it's safe to keep it (and curl-minimal)

build/Dockerfile
salv-orlando: versioning.mk was wrong, it was targeting the wrong version file.
I fixed versioning.mk and restored LDFLAGS in build.
So we now have the real version for the operator, and not anymore "0.0.3"

build/Dockerfile
salv-orlando: Done, and also added and e2e test case for version checks.

build/Dockerfile
salv-orlando: Yes, I had the .dockerignore, but I did not commit it.

test/e2e/operator_test.go
```diff
@@ -123,6 +123,29 @@ func TestOperatorE2E(t *testing.T) {
...
+		if err != nil || len(pods.Items) == 0 {
+			t.Fatalf("Failed to list operator pods: %v", err)
+		}
+
+		logBytes, err := clientset.CoreV1().Pods(operatorNamespace).
+			GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).
+			DoRaw(ctx)
+		if err != nil {
+			t.Fatalf("Failed to get operator pod logs: %v", err)
+		}
```
ggverma: No retry here — this is a single shot. Every other wait in this test suite wraps in `wait.PollImmediate`. The version line is logged at startup (`main.go:28`) before the readiness probe fires, so in practice it should be present by the time the deployment-ready subtest completes, but there's no guarantee from the API side about buffering/availability lag.

Suggested pattern (consistent with the rest of the suite):
```go
var logBytes []byte
err = wait.PollImmediate(2*time.Second, 30*time.Second, func() (bool, error) {
    logBytes, err = clientset.CoreV1().Pods(operatorNamespace).
        GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).
        DoRaw(ctx)
    if err != nil {
        return false, err
    }
    return strings.Contains(string(logBytes), expectedLine), nil
})
if err != nil {
    t.Fatalf(...)
}
```

ggverma:
**Design idea: expose version via `NcpInstallStatus` instead of logs**

The current log-based check works, but it couples the test to a specific log string format. A more durable alternative: add a `Version` field to `NcpInstallStatus` and have the operator write it there on startup (or when it sets the Available condition).

```go
// pkg/apis/operator/v1/ncpinstall_types.go
type NcpInstallStatus struct {
    Conditions []InstallCondition `json:"conditions,omitempty"`
    Version    string             `json:"version,omitempty"`
}
```

The status manager already calls `client.Get` on the NcpInstall CR and updates it — writing `version.Version` into `ncpInstall.Status.Version` would be a small addition. The e2e test could then read the CR status directly (the same pattern already used by the NCP rollout subtest) rather than grep-ing logs.

Benefits: no log format dependency, no buffering race, readable by any tool that can `kubectl get ncpinstall`.

This is a follow-on suggestion, not a blocker for this PR. What do you think — worth a separate change, or prefer to keep it in logs?

salv-orlando:
> **Design idea: expose version via `NcpInstallStatus` instead of logs**
> 
> The current log-based check works, but it couples the test to a specific log string format. A more durable alternative: add a `Version` field to `NcpInstallStatus` and have the operator write it there on startup (or when it sets the Available condition).

That's definetely worth doing, and I was already keen on making this change.
As It goes a bit beyond the scope of this PR, I will submit a new one, and will also update the test for leveraging status from NCPInstall rather than grepping the pod log.

test/e2e/operator_test.go
salv-orlando: Addressed.

ggverma (approved):

PR #310, opened 2026-06-29
fix: wire --zap-time-encoding flag via controller-runtime zap integra…

ggverma:
…tion

The operator was failing at startup with "unknown flag: --zap-time-encoding" because main.go used a hardcoded go.uber.org/zap logger that never registered the flag. Switch to controller-runtime's zap.Options.BindFlags() so all standard zap flags (including --zap-time-encoding=iso8601 in the deploy YAMLs) are properly recognised.


salv-orlando:
Thanks for the fix. I will amend the e2e tests to use the same yaml templates as production!

salv-orlando (approved):
LGTM

PR #314, opened 2026-07-15
Fix network status update failure with Server-Side Apply

xh013325:
## Summary
- Clear server-managed metadata fields (`managedFields`, `resourceVersion`, and `uid`) from the unstructured Network object before applying it.
- This prevents the Kubernetes API server from rejecting the Server-Side Apply request with a "metadata.managedFields must be nil" error.

## Test plan
- Verified that the codebase compiles and all unit tests pass successfully.

Made with [Cursor](https://cursor.com)


salv-orlando (approved):
LGTM

PR #316, opened 2026-07-20
Fix networkDiagnostics validation failure with Server-Side Apply

xh013325:
## Summary

This fixes an issue where the `nsx-ncp-operator` degrades on OpenShift due to an OpenAPI schema validation mismatch with newer OpenShift versions (OCP 4.16+).

Even though the `managedFields` issue was resolved, when the operator's configmap-controller (`updateNetworkStatus`) attempts to patch the `Network.config.openshift.io` resource, it serializes empty nested fields of the `networkDiagnostics` struct as JSON null values. The OpenShift API server's CRD validation strictly prohibits `null` for these fields (expecting objects or arrays).

To fix this, we remove the `spec` field from the apply configuration so that the operator only applies status/metadata changes, avoiding schema validation failures on uninitialized `spec.networkDiagnostics` fields.

## Test plan

- Confirmed `go build -o /dev/null ./cmd/manager` succeeds.

Made with [Cursor](https://cursor.com)


dantingl (approved):

heypnus (approved):

heypnus:
I had a quick chat with Xiao about this patch. Due to environment and time constraints, it hasn't been run through e2e tests yet. However, given its limited potential side effects, we decided to merge it first to speed up follow-up validation.

PR #321, opened 2026-08-05
DO NOT MERGE: E2E baseline check on main

ggverma:
Throwaway PR (empty commit, zero diff from main) to check whether the E2E test suite's `NcpInstall_CRD_Reconciliation_(Scaling_and_Tagging_Toggle)` failure seen in #320 reproduces on plain main, independent of that PR's changes. Will close after checking.


ggverma:
Baseline confirmed: E2E passes on plain main. Root cause and fix landed in #320.

PR #324, opened 2026-08-28
Fix error while updating Network.config.openshift.io

salv-orlando:
To avoid issues with spec missing the NetworkDiagnostics field, the code is currently entirely removing the spec field. This requires the ability to do server-side patch, which is not available in this branch.

This change works around this issue by submitting the Network's spec exactly as we retrieved it from the API server, since the only thing we want to update is the Network's status.

To this aim, we have to fetch the unstructured Network object, because the openshift API client used in this branch has an older version of the Network object which does not include fields added in OCP 4.16 and above such as NetworkDiagnostics.

### DIFF
--- a/pkg/controller/configmap/config.go
+++ b/pkg/controller/configmap/config.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


ggverma (approved):
Reviewed the diff and the BZticket root-cause writeup. The fix correctly preserves the original `spec` payload when updating Network status on release_4.2.5, since this branch's `apply.ApplyObject` uses a plain Get+Update (no server-side apply) unlike main/9.1.1 — stripping `spec` there sends an Update with no spec key and gets rejected by OCP 4.16+ validation. This targeted workaround is preferable to backporting the controller-runtime upgrade needed for real SSA. Verified in env per bug comments (nsx-ncp clusteroperator no longer degraded).

PR #326, opened 2026-09-18
OCP 4.22: ingress-canary network policy workaround

salv-orlando:
OpenShift's ingress-canary NetworkPolicy uses a namespaceSelector matching policy-group.network.openshift.io/host-network to allow host-network router pods to reach the canary. NCP translates namespaceSelectors into NSX group conditions on Segments, but host-network pods have no segment port at all - their IP is the node's IP - so NCP can never represent this selector. Traffic from the router pods is dropped, which fails the ingress operator's canary route check and blocks cluster install/upgrade on OCP 4.22.

Add a new, OpenShift-only controller that detects the upstream ingress-canary NetworkPolicy and, while it selects host-network namespaces, maintains a companion NetworkPolicy that explicitly allow-lists current node IPs as /32 ipBlocks, scoped to the canary pods and the canary Service's actual ports (discovered from the relevant service). The companion policy is owned by the upstream policy for automatic garbage collection.

Node and Service watches are predicate-filtered so only real IP or port changes trigger a reconcile, avoiding constant no-op re-applies from routine heartbeat/status updates. Each reconcile operation logs which watched kind and event type triggered it.

Testing done:
- Verified companion policy is created with correct node IPs, ports, pod selector and owner reference on a live OCP 4.22 cluster
- Verified ingress ClusterOperator's CanaryChecksSucceeding flips from failing to succeeding once the policy is applied
- Scaled the worker MachineSet up and down by one node twice and confirmed the new node's IP is added to/removed from the policy as soon as the node IP is available
- Confirmed no re-apply churn over several minutes of steady state after the node-address-change predicate fix
- Deleted the companion policy manually and confirmed it is recreated automatically
- Deleted the upstream ingress-canary policy and confirmed the companion policy is garbage-collected via its owner reference
- Verfied reconcile-trigger log lines correctly report kind, event type, namespace and name for Node, Service and NetworkPolicy watch events


pkg/controller/canaryworkaround/canaryworkaround_controller.go
```diff
@@ -0,0 +1,412 @@
...
+// canaryNetPol is set as the owner: the two policies are in the same
+// namespace (a hard requirement for owner references - cross-namespace
+// references are rejected by the garbage collector), and it means Kubernetes
+// automatically deletes the companion policy if canaryNetPol itself is ever
+// deleted, rather than relying solely on this controller noticing.
+func buildWorkaroundPolicy(canaryNetPol *networkingv1.NetworkPolicy, nodeIPs []string, ports []int32) *networkingv1ac.NetworkPolicyApplyConfiguration {
+	peers := make([]*networkingv1ac.NetworkPolicyPeerApplyConfiguration, 0, len(nodeIPs))
+	for _, ip := range nodeIPs {
+		peers = append(peers, networkingv1ac.NetworkPolicyPeer().
+			WithIPBlock(networkingv1ac.IPBlock().WithCIDR(fmt.Sprintf("%s/32", ip))))
```
dantingl: We do not support IPv6 on Openshift, right? do we need to filter out IPv6 address from listNodeIPs()

pkg/controller/canaryworkaround/canaryworkaround_controller.go
```diff
@@ -0,0 +1,412 @@
...
+		Name:      canaryServiceName,
+	}, svc)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get ingress-canary Service: %w", err)
+	}
+	ports := make([]int32, 0, len(svc.Spec.Ports))
+	for _, p := range svc.Spec.Ports {
```
dantingl: Should we check service targetPort? Or get port from NetworkPolicy

dantingl (dismissed):

deploy/openshift4/role.yaml
```diff
@@ -66,3 +66,7 @@ rules:
 - apiGroups: ['']
   resources: [events]
   verbs: [create, get, list, patch, update, delete]
+# Required by the canary-workaround controller create/delete a companion NetworkPolicy in openshift-ingress-canary.
+- apiGroups: [networking.k8s.io]
```
dantingl: Do we need to update bundle/openshift4/manifests/nsx-container-plugin-operator.clusterserviceversion.yaml

salv-orlando: That's a good point. Technically we don't because the CSV is used only when the operator is installed via the OLM. Since we don't certify the NCP operator, we can't install it via OLM, but we should not let the configuration drift, so I will update the CSV as well

pkg/controller/canaryworkaround/canaryworkaround_controller.go
salv-orlando: I am getting the ports from the Service, I would this it's the same as getting them from the network policy?

pkg/controller/canaryworkaround/canaryworkaround_controller.go
salv-orlando: We don't support IPv6, but I can filter out the IPv6 addresses.

pkg/controller/canaryworkaround/canaryworkaround_controller.go
salv-orlando: Maybe I misunderstood your comment. Yes, if targetPort is specified we should prefer it over port.

pkg/controller/canaryworkaround/canaryworkaround_controller.go
salv-orlando: Done

pkg/controller/canaryworkaround/canaryworkaround_controller.go
salv-orlando: Done

dantingl (approved):

PR #327, opened 2026-09-22
Fix nsx-ncp-operator rollout deadlock: change strategy to Recreate

salv-orlando:
When rolling out nsx-ncp-operator with replicas: 1, the default RollingUpdate strategy creates the new pod before tearing down the old one. With no pod anti-affinity rules, both pods can be scheduled on the same node, causing both to compete for the fixed metrics server port :8181. Since RollingUpdate waits for the new pod to be Ready (which it never becomes due to the port conflict), the rollout initially fails, until the old pod is removed. This causes a delay as in the meantime the new pod goes into CLBO.

Changing to Recreate strategy guarantees the old pod is fully torn down before the new one is created, eliminating any window where both can collide on the same node.


ggverma:
why not use soft pod anti affinity?

salv-orlando:
Because I want to make sure I can do a rollout even in the unlikely hypothesis that all the nodes but one are NotReady and notNotSchedulable

salv-orlando:
> Because I want to make sure I can do a rollout even in the unlikely hypothesis that all the nodes but one are NotReady and notNotSchedulable

I know you're going to reply: "I said SOFT anti-affinity"

